PR #74, opened 2025-09-05
Create TSDAE data

### LINKED ISSUE #49: Write a script that generates a file of TSDAE formatted data out of an extract of LOINC part descriptions.
- Use the spreadsheet Brady developed that maps LOINC codes to part descriptions. At this time, we don’t need to preserve the LOINC codes proper, we only care about the part descriptions.
- Read in the part descriptions and clean them (some have newlines in their text, others have multiple spaces between some of the words—get rid of all of that). Don’t apply any other scrambling or randomization. 
- Use Spacy to apply intelligent sentential tokenization (i.e. we can’t just split on periods since there are acronyms, we need smarter sentence breakups).
- Write each sentence of each part description into a text file, one sentence per line of the file.

bamader:
## Description

This changeset introduces the `tsdae.py` script, which will eventually contain all the code needed to run TSDAE pre-training. Right now, though, the script focuses on preparing the sentences data file used for TSDAE. The script reads in the LOINC part descriptions extract Brady prepared, and then applies a whole bunch of pre- and post-processing in conjunction with Spacy's sentence tokenizer to get things cleaned up and high-quality

## Related Issues

Closes #49 

## Additional Notes

I wasn't quite able to automate every cleaning operation needed to get the TSDAE file. After running the script, I hand-cleaned (yes I went through the whole file, I hate it thanks) something like 50 more bad sentences amidst the ~31k. I'm not sure if those 50 would have actually changed anything, but for LLMs quality representation is just so important I thought it worth mentioning. In practice, that means we should try to avoid re-running sentence file creation whenever possible so we don't lose those hand-filters (luckily, we shouldn't ever have to touch the creation code again).

## Checklist

## Checklist for Reviewers


src/dibbs_text_to_code/tsdae.py
```diff
@@ -0,0 +1,192 @@
+import re
+from typing import Union
+
+import spacy
+
+PART_DESCRIPTION_EXTRACTS_FILE = "snoinc_data_extracts/loinc_codes_with_part_descriptions.csv"
+OUTPUT_SENTENCES_FILE = "training_data_files/part_description_sentences.txt"
+
+
+def create_tsdae_data(nlp: spacy.language.Language, parts_fp: str, out_fp: str) -> None:
```
m-goggins, line 10: I was able to get it working locally when you change it to `spacy.Language`

m-goggins (commented):
Regex file looks good, thanks for making that change! Are you planning to add tests for the tsdae.py file?

bamader:
I wasn't planning on tests for the TSDAE code, since it's intended to be run once purely as something for data creation (and testing that a file gets created seems like testing just for the sake of having one). Honestly, saying that makes me think I might want to move the creation code to `/scripts` since it's not going to be called again. I could split that code off into `scripts` but then write some tests for the regex helpers since those could be cross-applicable? The structure of things is a little weird with this one and isn't meshing great with the more repeatable code modules structure we had with RL. Any preferences on the best way to structure things?

PR #75, opened 2025-09-05
Update Medical Terminology Extract Script

BradySkylight:
## Description

After discussing with Marcelle, we decided that it would be best if we only had one row per LOINC code and a column for each of the different term types (ie. Short Name, Display Name, Related Names, etc...).  I made this change, but in the process found that there were examples where one of the 'names'/'terms' for a code would also contain an ',' in it, so I changed the delimiter for our CSV files to be a '|' instead.  

While running the script I found that the HL7 Lab Interpretation URL was broken and returning a 404.  So, I found a better, more up-to-date, and stable URL to use to gather this data.  However, the structure of the returned result was completely different than the original, so I had to account for that in the code.

Overall the changes are still straightforward and simple, IMHO.

## Related Issues

None

## Additional Notes

None

## Checklist

- [X] I have ensured that the pull request is of a manageable size, allowing it to be reviewed within a single session.
- [X] I have reviewed my changes to ensure they are clear, concise, and well-documented.
- [X] I have updated the documentation, if applicable.
- [X] I have added or updated test cases to cover my changes, if applicable.
- [X] I have minimized the number of reviewers to include only those essential for the review.

## Checklist for Reviewers

### DIFF
--- a/scripts/terminology_valueset_sync.py
+++ b/scripts/terminology_valueset_sync.py
@@ -26,6 +26,9 @@ import os
 import sys
 
 import requests
+from dotenv import load_dotenv
+
+load_dotenv()
 
 # Set Terminology URLS
 LOINC_BASE_URL = "https://loinc.regenstrief.org/searchapi/loincs?"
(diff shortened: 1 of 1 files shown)


m-goggins (approved):
Thanks for making these changes, Brady!

PR #78, opened 2025-09-10
Update repo structure

m-goggins:
## Description

Updates the repo structure to move data curation and augmentation out of the `src` folder, as it does not need to be baked into the image.

Per our discussion in eng sync on 9/10/25, I also uploaded the results of `terminology_valyeset_sync.py` to the `data` folder so that we are all using the same data. We may wish to add schemas for each of the resulting datasets so that the shape of the data is known.

## Related Issues

n/a

## Additional Notes
Let me know if the READMEs are helpful or if you have any suggestions.


bamader (approved):
So much cleaner! I plan to add a `tuning` or `training` directory for my work on algorithm development stuff, once we get there--that should be okay with this structure, right?

PR #89, opened 2025-09-18
Add LOINC enhancement function

### LINKED ISSUE #64: Write a function that takes a LOINC code string as input and performs word-level enhancement via replacement, abbreviation, and acronym substitution

m-goggins:
## Description

Adds functionality to enhance LOINCs based on available abbreviations and/or synonyms. I also added some helper functions (path.py) that others may find useful.

## Related Issues

Closes #64 

One thing to note is that we only look for substring enhancements, e.g., more than 1 word, if no single word enhancements are possible. Further, we cannot guarantee that the minimum number of enhancements (`min_enhancements`) due to the data available from the LOINC API being fairly sparse.


data_curation/augmentation.py
```diff
@@ -1,7 +1,39 @@
...
+
+# TODO: move these loads to path or normalize module
+LOINC_METHOD_ENHANCEMENTS = path.read_json(
+    "data/snoinc_extracts/loinc_method_abbrv_syn_20250916.json"
+)
+LOINC_PROPERTY_ENHANCEMENTS = path.read_json(
+    "data/snoinc_extracts/loinc_property_abbrv_syn_20250916.json"
+)
+LOINC_SCALE_ENHANCEMENTS = path.read_json(
+    "data/snoinc_extracts/loinc_scale_abbrv_syn_20250916.json"
```
BradySkylight, line 19: Just realized that the files are still not completely static.  I just did an update to filter out some 'bad' results from this data.  Maybe we could change this to get the 'latest' file?  This may be a bit more complex....if anything, maybe we change this to be the files dated as of today (from my PR that 'should' be out shortly).

data_curation/augmentation.py
```diff
@@ -212,3 +244,130 @@ def insert_loinc_related_names(
...
+        - "synonyms": Replace words with semantically related terms.
+        - "all": Apply all of the above techniques.
+    :param max_enhancements: The maximum number of enhancements to apply.
+    :param min_enhancements: The minimum number of enhancements to apply.
+    :return: The enhanced text.
+    """
+    if max_enhancements <= min_enhancements:
+        raise ValueError("max_enhancements must be greater than min_enhancements")
+
+    words = [[word.lower().strip(), [i]] for i, word in enumerate(text.split())]
```
BradySkylight, line 270: This is a cool way to get the word and the indexes of the word within the text.  I could have used this on my deletion function.  This is great for cases where there is a word used more than once within a text.

utils/path.py
```diff
@@ -0,0 +1,36 @@
...
+
+def read_json(path: str) -> dict:
+    """
+    Loads a JSON file.
+    """
+    if not pathlib.Path(path).is_absolute():
+        # if path is relative, append to the project root
+        path = str(pathlib.Path(code_root(), path))
+    with open(path, "r") as fobj:
+        return json.load(fobj)
```
BradySkylight, line 36: I assume this file was added just to resolve the issues with pytest and being able to add tests for the data_curation section...is that correct?

data_curation/augmentation.py
```diff
@@ -212,3 +244,130 @@ def insert_loinc_related_names(
...
+    # Check for possible enhancements
+    possible_words_to_enhance = _check_for_enhancements(words)
+
+    # Choose number of enhancements to apply
+    # Look for substrings to enhance if there are no individual words to enhance
+    if len(possible_words_to_enhance) < 1:
+        words = _generate_substrings(words)
```
BradySkylight, line 277: I'm still not clear how we are generating the substrings here.

PR #90, opened 2025-09-26
Add UMLS Related names for all LOINC codes to a data curation file

### LINKED ISSUE #85: Create Function to pull relevant text that map to LOINC codes from UMLS

BradySkylight:
## Description

Using a LOINC code we can get related names from the UMLS API using the 'atom' query as well as the 'crosswalk' query, which will grab names from other terminologies where their code correlates to a LOINC code.  This process is very long (approx. 36 hours to complete).  I've added some fail safes so you can start/stop the process (or if it errors out) and pick up where you left off.  Every 500 LOINC codes being processed, the results from the UMLS queries will be written (append) to a partial file.  If there is an exception during the process it will write all the UMLS results into a partial file.  If you restart the process it will pull in all the results from the partial file and find the last LOINC code processed and will start there again.

If the process completes with no errors it will write out to a file in the snoinc extracts folder named: loinc_umls_related_names_(datetime).json.  

### Example:
"Zopiclone-N-oxide [Mass/volume] in Urine": {
        "code": "89043-4",
        "names": [
            "Zopiclone-N-oxide (U) [Mass/Vol]",
            "Zopiclone-N-oxide Ur-mCnc",
            "Zopiclone-N-oxide:MCnc:Pt:Urine:Qn",
            "Zopiclone-N-oxide:Mass Concentration:To identify measures at a point in time:Urine:Quantitative"
        ]
    },

## Related Issues

Closes #85 

## Additional Notes
- Found a bug where not ALL LOINC codes were being pulled, so I fixed it and reran ALL the processes to pull down the SNOINC files to have a full amount
- Previous LOINC files were missing 147 codes

## Checklist

- [X] I have ensured that the pull request is of a manageable size, allowing it to be reviewed within a single session.
- [X] I have reviewed my changes to ensure they are clear, concise, and well-documented.
- [X] I have updated the documentation, if applicable.
- [X] I have added or updated test cases to cover my changes, if applicable.
- [X] I have minimized the number of reviewers to include only those essential for the review.

## Checklist for Reviewers


BradySkylight:
> Nice work! The related names look very promising. This can be done as a separate step, but I imagine, we'll want the data as 1 row per LOINC, right @bamader?

I was wondering if I should have made this into a JSON file instead with a single key entry for each code and a list of names as the value.  I can do that, and process the existing file into that format as well.  Does that work better?  The number of columns will not be known, as it varies by code.

bamader (commented):
I think I might have been misunderstanding something earlier today when we synced after standup, so apologies for miscommunicating. Looking at things here, I do think we would benefit from a "consolidated" representation of LOINC codes, where we have one row in the file per unique LOINC code. We could accomplish this either through a JSON dict, or as we discussed, we could write things to a `.txt` file (right now I believe it's `.csv`) where each row corresponds to one code and includes both the "numeric" code (e.g. '1002132-4') and the "string name" of the code. Either way though I do agree with Marcelle's comment that we want to get down to one row per code.

bamader:
@BradySkylight Yep that works great--thanks for the refactor Brady!

PR #91, opened 2025-09-29
Add function to build augmented LOINC data files for testing

### LINKED ISSUE #62: Write a function that builds LOINC data set files
This function will create three augmented LOINC data files, one for each name variant (long common name, short name, and display name).
The function should accept as parameters a string path to a LOINC extract file, three augmentation configuration dictionaries (one for each name variant), and three numbers indicating how many samples of each name variant to generate (e.g. num_lcn, num_sn, num_dn)
Use the generate_augmented_loinc_data function from a previous ticket
For each name variant:

- Iterate through all LOINC codes in the sheet Brady pulled down, considering only code strings of that variant
- Run the generate_augmented_loinc_data function on this variant to get the appropriate number of constructed modifications
- Write the output into an “augmented_loinc_data_[variant_type].txt” file with the following format: a numeric standard LOINC code, a colon, a variant named LOINC code string, a colon, a pipe-delimited list of the 5 augmented strings generated for the original string. Here, variant_type is lcn, sn, dn
- Example: 
  - 12345-6:Textus to Codus LM:Txts to Cods LM|Codus to Textus Language Model| etc.
-

m-goggins:
## Description

This function creates 3 data files with augmented LOINC test data; 1 file for short names, 1 for long common names, and 1 for display names. 

## Related Issues

Closes #62


data_curation/augmentation.py
```diff
@@ -450,3 +450,61 @@ def generate_augmented_examples(
...
+    num_map = {"short_name": num_sn, "long_common_name": num_lcn, "display_name": num_dn}
+
+    # Read in data/loinc_lab_names.csv
+    with open(
+        input_path,
+        encoding="utf-8",
+    ) as fp:
+        data = fp.readlines()
+
+    for row in data[1:15]:
```
bamader, line 486: I'm guessing the subscripting is kruft?

data_curation/augmentation.py
```diff
@@ -450,3 +450,61 @@ def generate_augmented_examples(
...
+    :param num_lcn: The number of augmented long common names to generate.
+    :param num_sn: The number of augmented short common names to generate.
+    :param num_dn: The number of augmented display names to generate.
+    :param output_files_base: The base path for the output files.
+    :return: None
+    """
+
+    num_map = {"short_name": num_sn, "long_common_name": num_lcn, "display_name": num_dn}
+
+    # Read in data/loinc_lab_names.csv
```
bamader, line 479: Super tiny nit: with the directory re-org, this is now `data/snoinc_extracts/loinc_lab_names_XXXX.csv`

data_curation/augmentation.py
```diff
@@ -450,3 +450,61 @@ def generate_augmented_examples(
...
+        data = fp.readlines()
+
+    for row in data[1:15]:
+        r = row.split("|")
+        # skip any malformed rows
+        if len(r) < 5:
+            continue
+
+        loinc_code, short_name, long_name, display_name = r[0], r[1], r[2], r[3]
+        related_names = r[5].split(";") if r[4] else []
```
bamader, line 493: What's the logic for this conditional? Looking at the loinc_lab_names file, it looks like r[4] corresponds to the definition / description block (generally useless for us), and r[5] corresponds to related names as you have here. But why do we only split up the related names if the definition description exists? Theoretically it should be possible to have a code string like

```
code|long_common_name|short_name|display_name||rname_1;rname_2
```

where the definition entry is just a null, empty element in the `r` list, right?

bamader (commented):
Looks mostly good--just a kruft cleanup in indexing and a question on some related names conditional logic

data_curation/augmentation.py
```diff
@@ -476,21 +476,21 @@ def build_augmented_loinc_files(
...
         encoding="utf-8",
     ) as fp:
         data = fp.readlines()
 
-    for row in data[1:15]:
+    for row in data:
         r = row.split("|")
         # skip any malformed rows
-        if len(r) < 5:
+        if len(r) < 6:
```
BradySkylight, line 489: Were there cases in the data where they were malformed?

PR #125, opened 2025-10-30
Add embedding notebook

### LINKED ISSUE #114: SPIKE: Set up embedding script in Azure
Much like the evaluation script, it's not feasible to run the embedding generation locally anymore. It just takes too long. We need to get the infrastructure required for loading up a sentence transformers model and then computing embeddings deployed as an Azure notebook. The same considerations as applied to the evaluation notebook generally apply here, except that we will need an attached GPU for optimized computation. This is probably a harder ticket of work than the evaluation script, which isn't as time-consuming or resource intensive, so the following approach is suggested:

* Upload the relevant SNOINC extract that the mapped codestrings are in (`/snoinc_extracts/loinc_lab_names_20251008.csv`)
* Provision a small GPU instance (the cheapest one available, this is just beta testing) and attach it to a new notebook (I think we'll eventually want serverless, but for testing and development an attached compute cluster will probably be fastest for reporting times)
* Get the notebook to download a sentence transformers model into a local variable (using the `model = SentenceTransformers(...)` line in the script)
* Get the notebook to read the loinc lab names file and load that data up (can use the functions in the notebook, but might need to grab a copy of the `parse_snoinc` from utils)
* Try to compute an embedding for only a single vector to make sure the GPU is working
* Try to save that single vector into a pickle file and make sure the pickle file appears

Once everything is working on one example, we can process the whole file.

### LINKED ISSUE #131: Update embedding and performance notebooks with final changes
There were a few performance tweaks that needed to be made to the Azure notebooks for both the embedding and performance scripts. Once we're done collecting model performance data, we want to capture those in our repo. The scope of this ticket is to review the markdown documentation in each notebook for accuracy, cleaning up any outdated information and updating with new details, and then upload fresh copies of both notebook to the repo.

bamader:
## Description

This PR adds the embedding notebook from Azure into our repo. Not that anyone would try to run it locally, but just as a note it's unlikely to work outside of the cloud due to some `assert` statements I built in around `torch.cuda` (unless you have a really beefy GPU).

## Related Issues

Closes #114 
Closes #131 

## Additional Notes

Embedding with even the small crappy GPU that's super cheap is 20-30x faster than embedding on my local so we are _cooking_

### DIFF
--- a/azure_scripts/embedding.ipynb
+++ b/azure_scripts/embedding.ipynb
@@ -16,9 +16,9 @@
     "\n",
     "Note that for optimal performance, a GPU-supported compute instance is **required**. CPU-operations get Tensor-processed at a rate of ~1 iteration/second, while GPU-operations can be processed as fast as 30 iterations/second. \n",
     "\n",
-    "For larger models (e.g. those with many parameters), the notebook also supports mini-batching. This increases the time to compute embeddings but ensures the GPU isn't overloaded and no memory errors occur.\n",
+    "For larger models (e.g. those with many parameters), the notebook also supports mini-batching. This increases the time to compute embeddings but ensures the GPU isn't overloaded and no memory errors occur. For models of up to 8B parameters, mini-batching is not needed, and all vectors can be processed in memory. Any larger, and some mini-batching is required.\n",
     "\n",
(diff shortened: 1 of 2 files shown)


m-goggins (approved):
🚀

PR #127, opened 2025-10-31
feat: Front end components and styling

JNygaard-Skylight:
## Description

Adds the components and styling to the front end, with no logic and a dummy endpoint for the API.  Clicking "submit" returns a "good" result, with a single match. Clicking "bad" returns a "bad" result, where we do not have a high confidence match and instead return multiple possible codes.

## Related Issues

Closes

## Additional Notes
image
image
image
image

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -90,10 +90,10 @@ function App() {
               <div className="mt-4 flex">
                 <ButtonGroup type="segmented">
                   <Button type="button" onClick={handleSubmit}>
-                    Submit
+                    Submit <span className='text-xs'>(match)</span>
                   </Button>
                   <Button type="button" onClick={handleSubmitBad}>
-                    Bad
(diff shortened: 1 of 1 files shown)


BradySkylight (approved):
I think this looks good, and I think my only question is that the plan is to have only one button in the future for 'submit' and then depending on the results, the displays will correspond to the ones you have provided for 'good', 'bad', correct?  Nice work!

JNygaard-Skylight:
> I think this looks good, and I think my only question is that the plan is to have only one button in the future for 'submit' and then depending on the results, the displays will correspond to the ones you have provided for 'good', 'bad', correct? Nice work!

The multiple buttons are only there for development, on account of there not being any logic.

m-goggins (approved):
After talking with Josh, we're leaving as is until we have more feedback from users and/or better understand the outputs of the API.

PR #182, opened 2025-12-10
feature: Embed Service

### LINKED ISSUE #157: Write function to embed nonstandard text
### Description
Assuming the nonstandard text is of high enough quality, we need to embed (vectorize) it using the same model that we have used embed the LOINCs for the database.

```
model = SentenceTransformer(MODEL_NAME)
enc = model.encode(nonstandard_in)
```

### Acceptance Criteria
- [ ] A new function is added that has inputs for model name and nonstandard input; output is a vector
- [ ] Unit tests are included in the PR

JNygaard-Skylight:
## Description

Creates a simple "embed" service with a single function.

## Related Issues

Closes #157


src/dibbs_text_to_code/configs.py
```diff
@@ -21,3 +21,6 @@
     "lab_value": {},
     "lab_interp": {},
 }
+
+MODEL_NAME = "Qwen/Qwen3-Embedding-8B"
```
BradySkylight: Is this the model we landed on, or was it one of the 'snowflake' models?

src/dibbs_text_to_code/services/embed.py
```diff
@@ -0,0 +1,11 @@
...
+from torch import Tensor
+
+from dibbs_text_to_code.configs import MODEL_NAME
+
+model = SentenceTransformer(MODEL_NAME)
+
+
+def embed(input_text: str) -> Tensor:
+    """Embed text."""
+    return model.encode(input_text)
```
BradySkylight: This is awesome and simple!  Should we update the doc string to be a bit more descriptive?  This may only be relevant for maybe our actual APIs though....but just checking.

src/dibbs_text_to_code/configs.py
m-goggins: That's one of the models we're looking at, definitely fine as a placeholder until we figure out the final model

src/dibbs_text_to_code/services/extraction.py
```diff
@@ -0,0 +1,40 @@
+from dibbs_text_to_code import configs
+
+
+def _is_valid_data_field(data_field: str) -> bool:
+    return data_field.strip() in configs.DATA_FIELDS
+
+
+def _meets_word_count(text: str, word_count: int) -> bool:
+    return len(text.split()) > word_count
```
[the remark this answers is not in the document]
BradySkylight: That's my bad here actually.  Josh I can add them in another PR if you would like so you can just skip that for now.

src/dibbs_text_to_code/services/extraction.py
```diff
@@ -0,0 +1,40 @@
+from dibbs_text_to_code import configs
+
```
m-goggins: I do think you should add at least one test to show we've returned a tensor object that is the same dimensions as the model expects (768 for Qwen). I could be wrong, but I don't think you'll have to mock anything as long as you load this in the test file `from embeddings import embed`

BradySkylight: I like this idea as well!

src/dibbs_text_to_code/services/embed.py
```diff
@@ -0,0 +1,11 @@
+from sentence_transformers import SentenceTransformer
+from torch import Tensor
+
+from dibbs_text_to_code.configs import MODEL_NAME
+
+model = SentenceTransformer(MODEL_NAME)
```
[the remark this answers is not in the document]
BradySkylight: I added a couple of tests.

BradySkylight:
Just a few things to note....

1. I added a simple lazy load method, but we may need a different way to do this?
2. I had to change the model from Qwen to something more lightweight to get the tests to run in Github (as the Qwen model ran out of memory).  Locally the tests run, but it does take some time.
3. I adjusted the versions of the torch and added sentence-transformers to the dependencies - do we need this?  Is this too heavy?  ( I am assuming so...)  I'm calling it out so we can make adjustments in the future, though it is working for now.
4. I moved all functions into a single python file/package/service called text_processor, for everything dealing with processing/prepping text inputs for evaluation within the TTC module.  We can rename/adjust this as necessary.

I will add a ticket to the backlog, just one, to address all these questions/concerns later.

robertandremitchell (approved):
looks good to me. I wonder if we could get away with just mocking the model for test purposes, or if there's some way to cache it, but seems worth just rolling into the next ticket you already created.

PR #190, opened 2026-01-02
Get text candidates from an eICR message using the Schematron Error data as well as configured xpaths

### LINKED ISSUE #176: Get Possibly Relevant Strings
## Summary
For each relevant Schematron error we need to collect all possibly relevant strings that we may want to send to OpenSearch.

## Details

Once the previous ticket has been implemented we should have some kind of list of XPaths (that should point to either a planned observation or a result observation). For each XPath we need to collect the string values from the following locations (written as roughly compliant XPaths):
- `observation/code/@code`
- `observation/code/@displayName`
- `observation/code/originalText/text()`
- `observation/code/translation`
  - Note: translation is a list of codes, therefore for each translation you will have to look in all of the same places as the code above.
- `observation/text/text()`
There are potentially other places to look but these are the obvious places to start.

BradySkylight:
## Description

This PR adds a couple of functions that allow the TTC module to leverage the 'Base Xpath' provided by the Schematron output to extract a section from an eICR message and then, based upon what data field is the focus, the subsequent 'children xpaths', which are configured by data field, are used to extract text from various 'paths' within the eICR message section.  If the 'child xpath' finds text that is available, it stores it in a list that is returned. 

This PR also includes updates to the configs.py - adding the 'child Xpaths' for Lab Order and Lab Results.  There is also a helper function that adds the namespace to the xpaths.

~~NOTE: I had considered keeping the 'Base Xpath' and 'Child Xpath' together with the text in some kind of dictionary, but wasn't sure if that would be necessary.  Once we orchestrate all these functions together, there may be some refactoring to simplify the outputs etc....~~

**I have made the output of the function a dictionary where the base-xpath, sub-xpath, and iteration are the 'key' and the value is the actual text string in said 'key' xpath location.**

I added some tests to verify this functionality works on a sample eICR message I downloaded from the AIMs site and have included in our testing suite.

## Related Issues

Closes #176 

## Additional Notes

~~NOTE: I had considered keeping the 'Base Xpath' and 'Child Xpath' together with the text in some kind of dictionary, but wasn't sure if that would be necessary.  Once we orchestrate all these functions together, there may be some refactoring to simplify the outputs etc....~~

**I have made the output of the function a dictionary where the base-xpath, sub-xpath, and iteration are the 'key' and the value is the actual text string in said 'key' xpath location.**

## Checklist for Reviewers


src/dibbs_text_to_code/services/text_processor.py
```diff
@@ -126,3 +126,77 @@ def get_data_fields_from_schematron_error(schematron_output: str) -> dict:
             print(f"Error parsing schematron output: {e}")
```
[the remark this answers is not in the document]
m-goggins: The return could include some metadata that might be helpful for us later on when we evaluate how successful TTC is...something like the schematron sent 20 errors in this message, we had trouble parsing 1, fixed 17 of the remaining, and couldn't find a suitable replacement for the remaining 2. 

Now that I'm thinking it through, this should definitely be a separate ticket since it involves figuring out the schema for returning errors as well as metadata that may be useful for analysis.

src/dibbs_text_to_code/services/text_processor.py
```diff
@@ -126,3 +124,78 @@ def get_data_fields_from_schematron_error(schematron_output: str) -> dict:
...
+    :param data_field: The data field/element of interest for TTC processing.
+    :returns: A list of text candidates found within the eICR for
+        the specified data field/element for TTC processing.
+    """
+
+    text_candidates = []
+    namespaces = {"cda": "urn:hl7-org:v3"}
+    if (
+        eicr_data.strip() is None
+        or base_xpath.strip() is None
```
m-goggins:
```suggestion
       not  eicr_data.strip()
        or not base_xpath.strip()
```

if eicr_data = "", then `eicr_data.strip() is None` will still evaluate to False.

src/dibbs_text_to_code/configs.py
```diff
@@ -11,12 +11,26 @@
...
+            "/code/@displayName",
+            "/code/originalText/text()",
+            "/code/text/text()",
+            "/code/translation/@displayName",
+            "/code/translation/originalText/text()",
+            "/code/translation/text/text()",
+        ],
     },
     "lab_value": {},
     "lab_interp": {},
```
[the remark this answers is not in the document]
BradySkylight: Great idea.... I'm on it!

src/dibbs_text_to_code/services/text_processor.py
BradySkylight: It was failing in my unit tests and therefore I added that additional line...but typing would be a better way to tackle this for sure.

src/dibbs_text_to_code/services/eicr_processor.py
```diff
@@ -126,3 +126,77 @@ def get_data_fields_from_schematron_error(schematron_output: str) -> dict:
             print(f"Error parsing schematron output: {e}")
```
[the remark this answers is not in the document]
BradySkylight: Let's make this a separate ticket, but I'm leaving this comment here for now, so other reviewers can see the discussion/decision.

src/dibbs_text_to_code/configs/lab_order.py
```diff
@@ -0,0 +1,32 @@
...
+    Schema for Lab Order configuration settings
+    """
+
+    data_field: str = "lab_order"
+    """The data field/element this configuration applies to."""
+
+    text_word_count: int = 2
+    """The minimum word count required for text to be considered viable for TTC processing."""
+
+    xpaths: typing.List[str] = [
```
JNygaard-Skylight:
```suggestion
    xpaths: list[str] = [
```

src/dibbs_text_to_code/configs/lab_order.py
```diff
@@ -0,0 +1,32 @@
...
+        "/code/@displayName",
+        "/code/originalText/text()",
+        "/code/text/text()",
+        "/code/translation/@displayName",
+        "/code/translation/originalText/text()",
+        "/code/translation/text/text()",
+    ]
+    """The list of Sub XPath expressions to extract text in various locations from the lab order element."""
+
+    schematron_errors: typing.List[str] = [
```
JNygaard-Skylight:
```suggestion
    schematron_errors: list[str] = [
```

src/dibbs_text_to_code/configs/lab_value.py
```diff
@@ -0,0 +1,22 @@
...
+    Schema for Lab Value configuration settings
+    """
+
+    data_field: str = "lab_value"
+    """The data field/element this configuration applies to."""
+
+    text_word_count: int = 1
+    """The minimum word count required for text to be considered viable for TTC processing."""
+
+    xpaths: typing.List[str] = []
```
JNygaard-Skylight:
```suggestion
    xpaths: list[str] = []
```

src/dibbs_text_to_code/services/evaluator.py
```diff
@@ -0,0 +1,67 @@
...
+from sentence_transformers import SentenceTransformer
+from torch import Tensor
+
+from .utils import get_data_field_config
+
+MODEL: SentenceTransformer | None = None
+
+
+def _set_sentence_transformer(model_name: str = MODEL_NAME):
+    global MODEL
```
JNygaard-Skylight: `global` gives me the ick. Could we rewrite this to avoid using it?

src/dibbs_text_to_code/services/evaluator.py
```diff
@@ -0,0 +1,67 @@
+from configs.general import MODEL_NAME
+from sentence_transformers import SentenceTransformer
+from torch import Tensor
+
+from .utils import get_data_field_config
+
+MODEL: SentenceTransformer | None = None
+
+
+def _set_sentence_transformer(model_name: str = MODEL_NAME):
```
JNygaard-Skylight: Wouldn't this need to be public?

src/dibbs_text_to_code/services/utils.py
JNygaard-Skylight: These two functions are both related to getting the relevant config, so they could live in `configs` somewhere?

tests/unit/test_eicr_processor.py
```diff
@@ -0,0 +1,118 @@
+from pathlib import Path
+
+from services import eicr_processor
+
+current_dir = Path(__file__).parent
+
+
+class TestEICRProcessor:  # noqa: D101
```
JNygaard-Skylight: I do not think we typically need docstrings for test classes, so this could be added to the list of rules we ignore for tests.

tests/unit/test_eicr_processor.py
```diff
@@ -0,0 +1,118 @@
...
+    SCHEMATRON_ERROR_FILE = None
+    TEST_EICR_FILE = None
+
+    def file_setup(self):
+        if self.SCHEMATRON_ERROR_FILE is None:
+            schematron_path = current_dir / "assets" / "test_schematron_errors.xml"
+            with open(schematron_path, "r", encoding="utf-8") as f:
```
JNygaard-Skylight: [lxml can parse straight from a file path](https://stackoverflow.com/a/42901996), which I think would be cleaner.  E.g.: `lxml.parse(schematron_path)`

src/dibbs_text_to_code/services/eicr_processor.py
```diff
@@ -0,0 +1,128 @@
...
+                        data_fields_with_context[err_data_field] = []
+                    # if the xpath for a particular data field is already
+                    # accounted for, don't duplicate it
+                    if xpath not in data_fields_with_context[err_data_field]:
+                        data_fields_with_context[err_data_field].append(xpath)
+
+        except Exception as e:
+            # TODO: we may want to log this somewhere instead of print
+            print(f"Error parsing schematron output: {e}")
+            continue
```
m-goggins: I think it would be better to use the try/except more specifically, e.g., around the `get_data_field_by_schematron_error`, and instead of printing the error, include the error message in the dictionary that we return so you could return a dictionary with either the xpath or the error message. What do you think? Or do you think it makes sense to keep the errors separate from the xpaths we return?

tests/unit/test_eicr_processor.py
JNygaard-Skylight:
```suggestion
            with schematron_path.open() as f:
```

You've already created the `Path` object, so you can simplify this line.

src/dibbs_text_to_code/services/eicr_processor.py
```diff
@@ -0,0 +1,128 @@
...
+                for i, sub_node in enumerate(sub_nodes):
+                    if len(sub_node.strip()) > 0:
+                        # NOTE: I've added the itereator at the end of the key to ensure uniqueness
+                        # per key in the case that there may be multiple locations where the text
+                        # candidate may be the same
+                        text_candidates[base_xpath + sub_xpath + f"[{i}]"] = sub_node.strip()
+    except Exception as e:
+        # TODO: we may want to log this somewhere instead of print
+        print(f"Error extracting text from eicr message: {e}")
+        return text_candidates
```
m-goggins: Ditto the errors here; I think we should be more specific with the try/except block.

Unlike the try/except errors with the schematron output, I'm leaning towards not including them with the text_candidates but curious what you think.

src/dibbs_text_to_code/services/evaluator.py
```diff
@@ -0,0 +1,67 @@
...
+    global MODEL
+    if MODEL is None:
+        model = SentenceTransformer(model_name)
+        MODEL = model
+
+
+def embed(input_text: str) -> Tensor:
+    """Takes a text string and embeds it as vectorspip
```
m-goggins:
```suggestion
    """Takes a text string and embeds it as a vector
```

src/dibbs_text_to_code/services/utils.py
```diff
@@ -0,0 +1,33 @@
...
+        within focus, or None, for the TTC module.
+    """
+    try:
+        data_field_config_class = eicr_config[data_field.strip()].value.__class__
+        return data_field_config_class()
+    except KeyError:
+        return None
+
+
+def get_data_field_by_schematron_error(schematron_error: str) -> str | None:
```
m-goggins: Can you make this name more distinct from `get_data_fields_from_schematron_error`?

src/dibbs_text_to_code/configs/lab_interp.py
```diff
@@ -0,0 +1,22 @@
+# Configuration settings for lab interpretation processing in the TTC module
+import typing
+
+import pydantic
+
+
+class LabInterpConfig(pydantic.BaseModel):
```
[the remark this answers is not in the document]
BradySkylight: I was thinking about this - also in regards to Josh's comment to have a base class that inherits.  I'll clean this up.

src/dibbs_text_to_code/services/eicr_processor.py
```diff
@@ -126,3 +126,77 @@ def get_data_fields_from_schematron_error(schematron_output: str) -> dict:
             print(f"Error parsing schematron output: {e}")
             continue
     return data_fields_with_context
+
+
+def _enhance_xpath_with_namespace(xpath: str, namespace: str) -> str:
+    """Enhance an XPath with the specified namespace.
```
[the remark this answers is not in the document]
JNygaard-Skylight: I assume (hope) there's a better way to do this, but in the past the best I could figure out is aliasing the namespace to something shorter.

src/dibbs_text_to_code/services/eicr_processor.py
```diff
@@ -0,0 +1,128 @@
...
+            and part.endswith("()") is False
+        ):
+            enhanced_parts.append(f"{namespace}:{part}")
+        else:
+            enhanced_parts.append(part)
+    enhanced_xpath = "." + "/".join(enhanced_parts)
+    return enhanced_xpath
+
+
+def get_text_candidates(eicr_data: str, base_xpath: str, data_field: str) -> list:
```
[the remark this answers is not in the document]
BradySkylight: Added to a separate ticket #191

src/dibbs_text_to_code/services/evaluator.py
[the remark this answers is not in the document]
BradySkylight: Created separate ticket for this work. #192

src/dibbs_text_to_code/services/utils.py
BradySkylight: Done!

src/dibbs_text_to_code/configs/general.py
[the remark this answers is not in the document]
BradySkylight: I tried to make this work, but was unable to.  I'll include this in the class creation for configs ticket though, which it will make sense there and be easier to implement that way.

src/dibbs_text_to_code/services/utils.py
BradySkylight: Done!

src/dibbs_text_to_code/configs/general.py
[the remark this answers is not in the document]
BradySkylight: Added this to new ticket #193

src/dibbs_text_to_code/configs/__init__.py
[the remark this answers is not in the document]
BradySkylight: Added this to new ticket #193

src/dibbs_text_to_code/configs/lab_interp.py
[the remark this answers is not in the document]
BradySkylight: Added to new ticket #193

m-goggins (approved):
Thanks for making the changes and additional tickets @BradySkylight !

PR #222, opened 2026-01-22
chore: Update Dependabot config to be more chill

JNygaard-Skylight:
## Description

Updates the Dependabot config so that:
- It will check for updates for:
  - Github actions
  - Python packages
  - NPM packages
  - Docker images
- Will only open 2 PRs of each type at a time.
- Will check weekly on Wednesday morning so we do not get spammed during the work day.

## Related Issues

Closes n/a

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,26 +7,26 @@ updates:
     open-pull-requests-limit: 2
     schedule:
       interval: 'weekly'
-      day: 'monday'
+      day: 'wednesday'
       time: '12:00'
   - package-ecosystem: 'uv'
     directory: '/'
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,9 +1,26 @@
 version: 2
 updates:
-  - package-ecosystem: "pip"
-    directory: "/"
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    open-pull-requests-limit: 2
```
m-goggins: Does this mean it condenses all the fixes into two pull requests or would we only get two recommended fixes even if there are more than two high priority fixes?

PR #265, opened 2026-02-13
feat: Author element in header

JNygaard-Skylight:
## Description
Adds the ability to generate and insert an author tag into the augmented clinical document. 
```xml
<assignedAuthor>
    <id nullFlavor="NA"/>
    <addr nullFlavor="NA"/>
    <telecom nullFlavor="NA"/>
    <assignedAuthoringDevice>
        <softwareName displayName="Data Augmentation Tool"/>
    </assignedAuthoringDevice>
</assignedAuthor>
```
In addition this PR includes several refactorings:
- Splits `Augmenter` and `EICRAugmeneter` classes into seperate modules.
- Makes `Augmenter` an abstract base class.
- `Augmenter` is no longer a Pydantic model. As `Augmenter` is not a data model, this added a lot of complexity for no benefit.

## Related Issues

Closes #208

PR #276, opened 2026-02-19
feature: Add translations to augmentation

### LINKED ISSUE #210: Create Augmentation functionality to generate Translation Tag
### Basic Details & Requirements

- Create functionality within Augmentation Class to generate proper Translation Tags for TTC for Augmented eICRs
- Can be part of a 'Builder' function where it sets an additional attribute in Augmentation Class or we just create this as a function (getter function) that is called during orchestration to Augmenting an eICR
- Input - None - Use Class Attributes to get data/information necessary to generate tags- 
- We can follow and use a pattern from another DIBBs project - [example](https://drive.google.com/file/d/1AXkAxYyErjjCloIaP8_Shu3-Y-IbYsmc/view?usp=drive_link)

#### Translation Requirements and Examples:
- Create a new Translation tag with a sub <originalText> tag
   - Use the information from the TTC Output, per data element, - stored in the Augmentation Class as an attribute somehow - to set the <translation code= ... displayName= > attributes and the sub tag <originalText> </originalText>
      - code = Loinc Code returned by TTC for the data element
      - displayName = Loinc Text returned by TTC for the data element
      - originalText = Text that was evaluated by TTC which resulted in a found Loinc code/text
 
##### Rules & Example:

For additional information: https://docs.google.com/document/d/1dZSWvP3GWleU_JjE3w5IqxXBCqublGKB0xwf6yb5ycU/edit?tab=t.0

JNygaard-Skylight:
## Description

Adds translations to the augmented eICR:
```xml
<code displayName=" A custom code in display name. ">
    <originalText>
        A custom code in original text.
    </originalText>
    <translation code="10101010" codeSystem="2.16.840.1.113883.6.1" codeSystemName="LOINC" DisplayName="Chad new LOINC code" originalText="Loser old LOINC"/>
</code>
```

## Related Issues

Closes #210

## Additional Notes

This is built off [PR 274](https://github.com/CDCgov/dibbs-text-to-code/pull/274), which in turn is built off [PR 265](https://github.com/CDCgov/dibbs-text-to-code/pull/265). Review those first, please! :)


packages/augmentation/src/augmentation/models/config.py
```diff
@@ -45,6 +45,7 @@ class TTCAugmenterConfig(AugmenterConfig):
         ],
         DataField.LAB_TEST_NAME_RESULTED: [
             "author_entry",
+            "document_id_header",
```
BradySkylight: Should this be here?  If we have an over-arching document 'data field' then the document_id_header should be there and only added once.  If we do need to have document_id_header in this 'data field' then we will need it in all of them.

packages/augmentation/tests/unit/augmentor/snapshots/test_eicr_augmenter/test_basic_eicr/basic_eicr_augmented.xml
```diff
@@ -15,6 +15,7 @@
                                     <originalText>
                     A custom code in original text.
                   </originalText>
+                                    <translation code="10101010" codeSystem="2.16.840.1.113883.6.1" codeSystemName="LOINC" DisplayName="Chad new LOINC code" originalText="Loser old LOINC"/>
```
BradySkylight: the 'Chad' strikes again!  😆

PR #323, opened 2026-03-04
Add index creation lambda to ttc repo

### LINKED ISSUE #300: Add index creation lambda to ttc repo
### Description
This is a lambda function that runs one time during deployment to create the opensearch index with the appropriate mappings

### Acceptance Criteria

- [ ] Relevant code is ported over from dibbs-tf-envs, see code [here](https://github.com/skylight-hq/dibbs-tf-envs/blob/main/demo/text-to-code/lambda/index_lambda_function.py)
- [ ] Workflows are updated to make the index lambda publishable to GHCR
- [ ] Update the [index mappings](https://github.com/skylight-hq/dibbs-tf-envs/blob/main/demo/text-to-code/lambda/index_lambda_function.py#L58-L72) to accommodate the additional fields added in Brady's work: (#289):
  - id: string
  - description: string
  - loinc_name_type: string (searchable?)
  - description_vector: knn vector
  - type: string (searchable)
  - loinc_code: string (searchable)
  - property: string
  - time_aspect: string
  - system: string
  - scale_type: string
  - method_type: string
  - class_type: string
 - [ ] Consider making the mappings configurable rather than hard coding in the lambda

robertandremitchell:
## Description

Adds an initialization for the Dockerfile.index, lambda functions to create an index, updates the workflow to only run this lambda on first run / if changes detected.

Not sure if/how to store the mappings? I think based on Marcelle's notes on the ticket, I've updated correctly which fields from embeddings.ipynb that should be searchable. Not sure if this should be stored in a separate JSON, if we want to create a Class. Open to whatever!

## Related Issues

Closes #300 

## Additional Notes

## Checklist

## Checklist for Reviewers


packages/index-lambda/src/index_lambda/lambda_function.py
```diff
@@ -0,0 +1,130 @@
...
+                "loinc_type": {"type": "text"},
+                "loinc_code": {"type": "text"},
+                "loinc_name_type": {"type": "text"},
+                "property": {"type": "keyword"},
+                "time_aspect": {"type": "keyword"},
+                "system": {"type": "keyword"},
+                "scale_type": {"type": "keyword"},
+                "method_type": {"type": "keyword"},
+                "class_type": {"type": "keyword"},
+                "type": {"type": "text"},
```
BradySkylight: I think this can be removed.  We only have 'loinc_type' (formally known as 'type' in the split files), 'loinc_name_type' and the other axis types (class, method, scale).

Dockerfile.index
```diff
@@ -0,0 +1,27 @@
...
+
+# Initialize the index_lambda directory
+RUN mkdir -p "${LAMBDA_TASK_ROOT}/src/index_lambda"
+
+# Install build tools for compiling C++ extensions
+RUN microdnf install -y gcc-c++ make && microdnf clean all
+
+# Copy over just the pyproject.toml file and install the dependencies doing this
+# before copying the rest of the code allows for caching of the dependencies
+COPY ./packages/index-lambda ${LAMBDA_TASK_ROOT}/index-lambda
```
m-goggins: You'll also want to copy over `text-to-code-lambda/src/s3_handler.py to make sure we have those functions available. I think we _could_ split out s3_handler into it's own package since we'll also re-use the functions in augmentation but it doesn't have to be in this PR and we could always just copy directly from ttc-lambda.

packages/index-lambda/src/index_lambda/lambda_function.py
m-goggins: This is my bad in the original code, but after looking at the [docs](https://docs.opensearch.org/latest/mappings/supported-field-types/string/) a bit more, I think all of the fields that are `text` should be `keyword` instead, except `description` should remain `text` (and `description_vector` should stay the same).

Keeping the majority of the fields as keywords will make the queries more efficient because we'd only ever want to filter on exact matches for all of the fields except the `description`.

packages/index-lambda/src/index_lambda/s3_handler.py
```diff
@@ -0,0 +1,162 @@
...
+    :param object_key: The S3 object key
+    :param input_prefix: The pipeline step prefix (e.g., "TTCInput/")
+    :return: The persistence_id portion of the key
+
+    """
+    if not object_key.startswith(input_prefix):
+        raise ValueError(
+            f"Object key '{object_key}' does not start with expected prefix '{input_prefix}'"
+        )
+    return object_key[len(input_prefix) :]
```
m-goggins: Thanks for making the ticket! I think if you take on the ticket right away, this is fine. Otherwise, I'd prefer to just import from the text-to-code-lambda folder rather than duplicate the code

packages/index-lambda/src/index_lambda/s3_handler.py
```diff
@@ -0,0 +1,162 @@
+import os
```
m-goggins: Thank you for making the ticket! I think you should just import this functionality from text-to-code-lambda/s3_handler rather than duplicating the code (at least until the new ticket gets addressed)

PR #352, opened 2026-03-17
test: Sort data curation JSONs

JNygaard-Skylight:
Sorts JSONs being read by the data_curation package to get around inconsistent file reading across OSs, which then causes to unit tests to fail, even with a set random seed.. The more correct solution would be to mock `random.choice`, for every unit test, but this is significantly easier, and has no overhead because are only only a hand few of JSONs.


bamader (approved):
Great catch and thanks for the quick fix!

PR #365, opened 2026-03-23
Add Reranker service to lambda

### LINKED ISSUE #307: Write lambda functions that will invoke the model for embedding, querying, and reranking
Once we have a model newly tuned, and once that model is "available" to the Lambda, we need to be able to invoke functions of the model from within the Lambda itself. The scope of this ticket is to write functions within the lambda container that will apply functions from the model's package to incoming input. The functions we need to write are:

- encode_and_query, which invokes the model.encode() function to vectorize input and then send it to open search
- rerank_query_results, which invokes a reranker on the results returned from OpenSearch

bamader:
## Description

This PR adds the Reranker service into the TTC repo and to the lambda function handler. It creates the Reranker as a separate service class, following the pattern of the Embedder, and incorporates reranking into the record processing pipeline defined in the text to code lambda function. 

It also makes a small change to the index creation variables in the index lambda to use the EF Construction and M-Value parameters we experimentally determined were best for performance.

## Related Issues

Closes #307 

## Additional Notes

## Checklist

## Checklist for Reviewers

### DIFF
--- a/Dockerfile.ttc
+++ b/Dockerfile.ttc
@@ -24,7 +24,7 @@ RUN python -c "\
     from huggingface_hub import snapshot_download; \
     snapshot_download( \
     repo_id='intfloat/e5-large-v2', \
-    local_dir='/opt/model', \
+    local_dir='/opt/retriever_model', \
     ignore_patterns=['*.git*', '*.md', 'onnx/*', 'openvino/*', 'pytorch_model.bin'] \
     )"
 
@@ -33,10 +33,11 @@ RUN python -c "\
     from huggingface_hub import snapshot_download; \
     snapshot_download( \
     repo_id='cross-encoder/stsb-roberta-large', \
-    local_dir='/opt/model', \
+    local_dir='/opt/reranker_model', \
     ignore_patterns=['*.git*', '*.md', 'onnx/*', 'openvino/*', 'pytorch_model.bin'] \
     )"
 
-ENV MODEL_PATH="/opt/model"
+ENV RETRIEVER_MODEL_PATH="/opt/retriever_model"
+ENV RERANKER_MODEL_PATH="/opt/reranker_model"
 
 CMD ["text_to_code_lambda.lambda_function.handler"]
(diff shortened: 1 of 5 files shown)


Dockerfile.ttc
```diff
@@ -19,14 +19,23 @@ RUN pip install --no-cache-dir "${LAMBDA_TASK_ROOT}/text-to-code"
...
-)"
+    )"
+
+# Download reranker at build time (baked into image)
+RUN python -c "\
+    from huggingface_hub import snapshot_download; \
+    snapshot_download( \
+    repo_id='cross-encoder/stsb-roberta-large', \
+    local_dir='/opt/model', \
+    ignore_patterns=['*.git*', '*.md', 'onnx/*', 'openvino/*', 'pytorch_model.bin'] \
```
nickclyde: Both the retriever and reranker are downloaded to `/opt/model`. The second `snapshot_download` will overwrite files from the first, leaving only the reranker model on disk.

They need separate directories, e.g.:
```dockerfile
local_dir='/opt/retriever-model'
...
local_dir='/opt/reranker-model'
```

The `MODEL_PATH` env var also only points to one path — we'll need a second env var (e.g. `RERANKER_MODEL_PATH`) so each service can locate its own model at runtime.

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -264,8 +265,16 @@ def _process_schematron_errors(
...
         )
 
+        # The OpenSearch results object has a couple levels of nesting,
+        # but all we care about for reranking is extracting the actual
+        # text strings of the ANN LOINC codes
+        results_list = opensearch_retrieved_scores.hits.hits
+        retrieved_loinc_names = [hit.source.description for hit in results_list]
+        ranked_results = reranker.Reranker().rerank(selected_candidate.value, retrieved_loinc_names)
+
         metadata_error = error.model_dump()
```
nickclyde: Nit: `Reranker()` is instantiated on every iteration of the loop. While the heavy `CrossEncoder` is a module-level singleton so this is cheap, creating a new wrapper object each time is unnecessary. Try instantiating once before the loop (or at module level)

packages/index-lambda/src/index_lambda/lambda_function.py
```diff
@@ -21,7 +21,7 @@ def handler(event: dict, context: dict) -> dict:
                         "name": "hnsw",
                         "space_type": "cosinesimil",
                         "engine": "faiss",
-                        "parameters": {"ef_construction": 128, "m": 16},
+                        "parameters": {"ef_construction": 400, "m": 64},
```
nickclyde: Are we good to boost these? Increasing m increases memory per node, right? Should we check against the Lambda/OpenSearch resource limits? We're using r5.large.search

Issue #412, opened 2026-04-01
Fix Caching of Clients in Lambda


JNygaard-Skylight:
**Describe the bug**
We had initially decided that clients (S3, auth, OpenSearch) should be cached so that they persist over lambda runs. This is not happening. Instead they get initialized twice. Once at the beginning of lambda function, and once again when they are used. Compared to the rest of the code the initilizaing of these clients is pretty quick so it may be most prudent to simply remove this faulty logic. However if it would not take too much effort there is nothing in wrong with caching elements that are consistent across runs.

**Impact**
Please describe the impact this bug is causing to your program or organization.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Logs**
If applicable, please attach logs to help describe your problem.

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #429, opened 2026-04-06
test: Local e2e Tests

### LINKED ISSUE #378: E2E test harness for Lambda pipeline (moto-based)
## Description

Create a local E2E test harness using moto to mock AWS services, exercising the full pipeline:

1. Upload a test eICR + Schematron error file to mock S3
2. Invoke the TTC Lambda with a synthetic SQS event
3. Verify TTC output is written to the correct S3 prefix with expected structure
4. Invoke the Augmentation Lambda with the TTC output
5. Verify the augmented eICR is valid XML with expected `<translation>` elements

## Acceptance Criteria
- [ ] Pytest-based test suite in a `tests/e2e/` directory
- [ ] Uses moto for S3 and mocked OpenSearch
- [ ] Canonical test fixtures: sample eICR, Schematron errors, expected outputs
- [ ] Tests run in CI (no real AWS credentials needed)
- [ ] Documents what the E2E tests cover vs. what they don't

## Depends On
- #367 (remove hardcoded candidate)
- #213
- #263

JNygaard-Skylight:
## Description

- Local e2e testing. 
- Environment variables required for testing can now be kept in `.env.test`.

To run e2e tests:
```
just test -m e2e
```

To run all but the e2 tests:
```
just test -m \"not e2e\"
```

## Additional Comments
Running all tests at once, as in `just test` will cause some tests to fail I'm assuming because multiple tests are trying to use boto3 at the sametime. There should be a way to isolate the tests, and to clean up after each test, but I think running then separately is reasonable for now.

For reasons I don't quite understanding some pre-commit formatting was triggered in some non-relevant files. Probably because of some dubious git commands I ran to revert a lot of my changes.

Resolves #378


e2e/conftest.py
```diff
@@ -0,0 +1,115 @@
...
+from shared_models import TTC_METADATA_PREFIX
+from shared_models import TTC_OUTPUT_PREFIX
+from text_to_code_lambda import lambda_function
+
+AWS_REGION = "us-east-1"
+AWS_ACCESS_KEY_ID = "test_access_key_id"
+AWS_SECRET_ACCESS_KEY = "test_secret_access_key"  # noqa: S105
+OPENSEARCH_ENDPOINT_URL = "https://test-opensearch-endpoint.com"
+TEST_PERSISTENCE_ID = "2025/09/03/1-5f84c7a5-91d7f5c6a2b7c9e08f0d1234"
+
```
robertandremitchell: I think these are all or should be in `test_constants.py`

packages/lambda-handler/src/lambda_handler/lambda_handler.py
```diff
@@ -116,7 +135,7 @@ def get_eventbridge_data_from_s3_event(event: lambda_events.EventBridgeEvent) ->
 
 
 def put_file(
-    file_obj: typing.BinaryIO,
+    file_obj: any,  # ty:ignore[invalid-type-form]
```
robertandremitchell: we should know file_obj, right?

packages/lambda-handler/src/lambda_handler/lambda_handler.py
```diff
@@ -103,6 +102,26 @@ def get_file_content_from_s3(
...
+    bucket_name: str, object_key: str, s3_client: BaseClient | None = None
+) -> dict:
+    """Extracts the file content from an S3 bucket.
+
+    :param bucket_name: The name of the S3 bucket.
+    :param object_key: The key of the S3 object.
+    :param s3_client: Optional pre-created S3 client. If None, a new client is created.
+    :return: The content of the file as a string.
+    """
+    client = s3_client or create_s3_client()
```
robertandremitchell: should change based on #412

robertandremitchell (commented):
looks really awesome! know it's still under draft. looks like there's a handful of lines that aren't covered.

Also, not sure if it's work for this ticket, but i got on error on `typing` that was resolved with `from typing_extensions import TypedDict` for TypedDict. It looks like some of this changed in 3.12 (my virtualenv is currently 3.11). uv.lock looks like it ought to support 3.11 in some places, but not others. Not sure if we want to just upgrade fully to 3.12, which is fine, or switch the two instances of TypedDict to that type of import that is supported by 3.11.

Very minor, but there's a handful of lines in the coverage report also uncovered, that could probably just be addressed in a future ticket.

Will give a second look once it's fully ready for review, but because there was so much I wanted to start in a little early, hope that's okay/helpful.

packages/lambda-handler/src/lambda_handler/lambda_handler.py
JNygaard-Skylight: I think I resolved this?

pyproject.toml
```diff
@@ -30,10 +36,14 @@ dev = [
 source = ["packages"]
 omit = ["*/__pycache__/*", "*/test_*.py", "*/tests/*"]
 
-[tool.pytest.ini_options]
-testpaths = ["packages"]
+[tool.pytest]
+markers = ["e2e"]
+testpaths = ["e2e", "packages"]
 pythonpath = ["."]
```
nickclyde: Not sure if this is what's causing my failure, but does this section needs to be `[tool.pytest.ini_options]`? I'm seeing some docs say that pytest reads its pyproject config from that instead of `[tool.pytest]`

pyproject.toml
```diff
@@ -13,12 +13,18 @@ authors = [
 ]
 license-files = ["LICENSE"]
 readme = { file = "README.md", content-type = "text/markdown" }
-requires-python = ">=3.11,<3.13"
+requires-python = "==3.12.*"
```
nickclyde: Do we really need to narrow this to `==3.12.*`? You said it's for `TypedDict`, but `TypedDict` has been in `typing_extensions` since Python 3.8, so it should work fine on 3.11. Narrowing also locks out 3.13, which we may want in the future. Could we maybe widen this back to `>=3.11,<3.13` (or `>=3.12,<3.14`)?

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -106,57 +106,31 @@ def process_record(record: SQSRecord, s3_client: BaseClient, opensearch_client:
     # Parse the EventBridge S3 event from the SQS message body
     eventbridge_data = lambda_handler.get_eventbridge_data_from_s3_event(s3_event)
     object_key = eventbridge_data["object_key"]
-    bucket_name = eventbridge_data.get("bucket_name") or S3_BUCKET
+    bucket_name = eventbridge_data.get("bucket_name") or _S3_BUCKET
     logger.info(f"Processing S3 Object: s3://{bucket_name}/{object_key}")
```
nickclyde: Heads up: this computes `bucket_name` from the event and logs it, but then the rest of the pipeline (the new `_load_schematron_data_fields`, `_load_original_eicr`, and `_save_ttc_outputs`) all read `_S3_BUCKET` directly from the module global rather than using `bucket_name`. The net effect is that `bucket_name` is dead after this log line.

This is going to conflict with PR #438, which removes the `S3_BUCKET` env var entirely and relies on exactly this event-derived `bucket_name` being plumbed through the whole pipeline so one Lambda can serve both the production eCR bucket and the `ecr-ttc-training` bucket. Whichever PR merges second will need to re-plumb `bucket_name` through these functions — I'd rather we do it here so #438 doesn't silently regress when this lands on top of it.

nickclyde (commented):
Thanks for setting up end-to-end tests Josh!! Having a real pipeline walkthrough is going to be super useful. I do want to flag some concerns before merging, though, because the PR as it stands is doing a lot and I think a few of the non-e2e changes have introduced bugs. On first read through, these are the changes I'm noticing (writing them down here to keep track during my review):

The PR description mentions the e2e tests, an `.env.test` file, and pinning Python to 3.12, but the diff also:
- Renames `NonstandardCodeInstance` > `NonstandardCodeReplacement` and drops its `schematron_error` field
- Replaces the `TTCOutput`/`TTC metadata` dict-shaped payloads with pydantic models
- Moves `OpenSearchResult`/`Candidate`/`LabXPaths`/`SchematronErrorDetail` from package-local modules into `shared_models`
- Removes `require_env` from `lambda-handler` in favor of a new `utils.get_env_var`
- Changes `EICRAugmenter` to increment the original `versionNumber` rather than hard-coding `1`
- Deletes several helper functions (`_initialize_ttc_outputs`, `_populate_eicr_metadata`, `_build_nonstandard_code_instance`, `_parse_nonstandard_codes`) and the tests that covered them
- Removes the `assigning_authority_name`, `displayable`, and `null_flavor` fields on `CdaInstanceIdentifier`

This is a lot of changes! Which is totally fine, but it would be helpful to understand your reasoning for these changes. Right now the +1890/-1966 size makes it very hard to reason about what's intentional.

Also: PR #438 removes the `S3_BUCKET` env var entirely and derives the bucket from the S3 event payload so one Lambda can serve multiple buckets (production eCR vs training). This PR is still relying on a hardcoded `_S3_BUCKET` throughout both lambdas — something to consider, but see inline comments below.

As a last note, I'm not able to run the tests locally, I'm getting:

```
❯ uv run pytest -m e2e
ImportError while loading conftest '/Users/nick/Code/dibbs-text-to-code/e2e/conftest.py'.
e2e/conftest.py:6: in <module>
    from text_to_code_lambda import lambda_function
E   ModuleNotFoundError: No module named 'text_to_code_lambda'
```

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -106,57 +106,31 @@ def process_record(record: SQSRecord, s3_client: BaseClient, opensearch_client:
...
 
     :param persistence_id: The persistence ID extracted from the S3 object key
     :param s3_client: The S3 client to use for fetching files.
     :param bucket_name: The S3 bucket name to read from.
     :return: The relevant Schematron data fields for TTC processing.
     """
-    object_key = f"{SCHEMATRON_ERROR_PREFIX}{persistence_id}"
-    logger.info("Loading Schematron errors", s3_key=f"s3://{bucket_name}/{object_key}")
+    object_key = f"{_SCHEMATRON_ERROR_PREFIX}{persistence_id}"
+    logger.info("Loading Schematron errors", s3_key=f"s3://{_S3_BUCKET}/{object_key}")
```
nickclyde: `_load_schematron_data_fields` dropped its `bucket_name` parameter and now hard-codes `_S3_BUCKET`, same for `_load_original_eicr` and `_save_ttc_outputs` below. Same concern as my comment on `process_record`: this regresses the source-bucket routing from #438 and will need to be reverted once that lands. Could we keep `bucket_name` as a parameter and plumb the event-derived value through?

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -290,53 +208,39 @@ def _process_schematron_errors(
...
+            query=query, index=_OPENSEARCH_INDEX, opensearch_client=opensearch_client
         )
 
         # The OpenSearch results object has a couple levels of nesting,
         # but all we care about for reranking is extracting the actual
         # text strings of the ANN LOINC codes
         results_list = opensearch_retrieved_scores.hits.hits
         retrieved_loinc_names = [hit.source.description for hit in results_list]
+
         ranked_results = RERANKER.rerank(selected_candidate.value, retrieved_loinc_names)
```
nickclyde, line 313: I think this path crashes when OpenSearch returns zero hits. If `results_list` is empty then `ranked_results` is also empty (or the dict comprehension is empty), and `ranked_results[0]` / the subsequent `[...]` lookup will raise. The previous version explicitly handled the `not results_list` branch by writing an `unmatched_schematron_errors` entry. Could we keep an empty-result guard so a schematron error with no KNN match doesn't fail the whole record?

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -390,43 +293,62 @@ def _process_record_pipeline(
...
+    ttc_output = TTCOutput(
+        message=None,
+        persistance_id=persistence_id,
+        nonstandard_codes=[x.to_nonstandard_code_replacement() for x in schematron_errors],
+    )
+
+    ttc_metadata = TTCMetadata(
+        persistance_id=persistence_id,
+        message=None,
+        eicr_metadata=EicrProcessor(original_eicr_content).eicr_metadata,
```
nickclyde: Minor perf nit: `_process_schematron_errors` constructs a new `EicrProcessor(original_eicr_content)` inside the loop on every schematron error, and then here we construct yet another one just to read `eicr_metadata`. The eICR XML is reparsed N+1 times per record. Would it be simpler to build the `EicrProcessor` once in `_process_record_pipeline` and pass it in?

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -83,55 +82,50 @@ def _process_record(record: SQSRecord, s3_client: BaseClient) -> None:
...
-    ttc_output = _load_ttc_output(persistence_id, s3_client, bucket_name)
-    original_eicr = _load_original_eicr(persistence_id, s3_client, bucket_name)
-    nonstandard_codes = _parse_nonstandard_codes(ttc_output)
-
-    augmenter_input = TTCAugmenterInput(
-        eicr_id=persistence_id,
-        nonstandard_codes=nonstandard_codes,
-    )
+    ttc_output = _load_ttc_output(persistence_id, s3_client, _S3_BUCKET)
+    original_eicr = _load_original_eicr(persistence_id, s3_client, _S3_BUCKET)
```
nickclyde: I think there's an inconsistency here: `bucket_name` is computed from the event, but then `_load_ttc_output` and `_load_original_eicr` are called with `_S3_BUCKET` while `_save_augmentation_outputs` (further down) is still called with `bucket_name`. Also the log line uses `_S3_BUCKET` rather than the `bucket_name` the line above it just computed.

Once PR #438 is in place (and there's no `S3_BUCKET` env var at all), the reads here would fail while the writes would succeed to a different bucket. Could we pass `bucket_name` consistently to all three helpers and use it in the log line?

packages/lambda-handler/src/lambda_handler/lambda_handler.py
```diff
@@ -152,7 +141,9 @@ def put_file(
     :param s3_client: Optional pre-created S3 client. If None, a new client is created.
     """
     client = s3_client or create_s3_client()
-    client.put_object(Body=file_obj, Bucket=bucket_name, Key=object_key)
+    client.put_object(
+        Body=file_obj, Bucket=bucket_name, Key=object_key, ContentType="application/json"
```
nickclyde: I believe setting `ContentType="application/json"` here is a bug — the augmentation lambda uses this same helper to put the augmented eICR XML (see `_save_augmentation_outputs` in `augmentation_lambda/lambda_function.py`), so we'd be labeling XML objects as `application/json` in S3.

Could we either accept a `content_type` parameter on `put_file` (defaulting to not setting it) or move the ContentType into the caller that actually writes JSON?

packages/text-to-code/src/text_to_code/services/eicr_processor.py
```diff
@@ -155,8 +152,13 @@ def _create_xml_tree(xml: str) -> Element:
     tree = etree.fromstring(xml.encode("utf-8"))
     for elem in tree.iter():
         # Remove namespace from tag
-        elem.tag = etree.QName(elem).localname
+        try:
+            elem.tag = etree.QName(elem).localname
+        except Exception:
+            print(elem.tag)
```
nickclyde: This swallows any exception during tag localname extraction and falls back to `print(elem.tag)`, which I think is problematic for a couple of reasons:
1. `print` inside a Lambda goes to stdout without structured fields; we should use the `logger` (probably `logger.warning`).
2. Catching bare `Exception` and continuing hides the bug rather than fixing it. What was the original failure that motivated this?

packages/text-to-code-lambda/tests/conftest.py
```diff
@@ -127,9 +113,52 @@ def mock_aws_setup(monkeypatch: pytest.MonkeyPatch) -> boto3.client:
...
+        )
+        s3.create_bucket(Bucket=S3_BUCKET)
+
+        # Add convenience attributes for tests
+        s3.bucket_name = S3_BUCKET
+        s3.persistence_id = TEST_PERSISTENCE_ID
+
+        # Put test Schematron error file in the mock S3 bucket
+        schematron_path = Path(
+            "/Users/jnygaard/Dev/Skylight/Dibbs/dibbs-text-to-code/e2e/assets/test_schematron_errors.xml"
```
nickclyde: I'm sure this was just a quick hack you missed and forgot to clean up, but these absolute paths (`/Users/jnygaard/Dev/Skylight/Dibbs/dibbs-text-to-code/...`) won't exist on anyone else's machine or in CI — could we swap these for `Path(__file__).parent / ... ` or the project-root-relative paths used in `mock_aws_setup` right above?

packages/shared-models/src/shared_models/__init__.py
```diff
@@ -59,12 +56,147 @@ class NonstandardCodeInstance(BaseModel):
...
+
+class TTCOutput(BaseModel):
     model_config = ConfigDict(
         frozen=True,
         extra="forbid",
     )
-    eicr_id: str
-    nonstandard_codes: list[NonstandardCodeInstance]
+    message: str | None
+    persistance_id: str
```
nickclyde: I think `persistance_id` is a typo for `persistence_id` — same spelling as `TTCMetadata` below. These show up in the saved JSON on S3 (see the new snapshot fixtures), so let's fix before anything downstream depends on it would save a migration later.

e2e/test_e2e.py
```diff
@@ -0,0 +1,282 @@
...
+
+from augmentation_lambda.lambda_function import handler as augmentation_lambda
+from text_to_code_lambda.lambda_function import handler as ttc_handler
+from utils import get_env_var
+
+AUGMENTATION_METADATA_PREFIX = get_env_var("AUGMENTATION_METADATA_PREFIX")
+AUGMENTED_EICR_PREFIX = get_env_var("AUGMENTED_EICR_PREFIX")
+AWS_ACCESS_KEY_ID = get_env_var("AWS_ACCESS_KEY_ID")
+AWS_SECRET_ACCESS_KEY = get_env_var("AWS_SECRET_ACCESS_KEY")
+AWS_ACCESS_KEY_ID = "test_access_key_id"
```
nickclyde: `AWS_ACCESS_KEY_ID` is assigned twice — the literal `"test_access_key_id"` on line 22 shadows the `get_env_var("AWS_ACCESS_KEY_ID")` result from line 20. I'm guessing the second assignment is a leftover from iterating on this file and can be deleted.

.env.test
```diff
@@ -0,0 +1,19 @@
...
+SCHEMATRON_ERROR_PREFIX = "schematronErrors/"
+TTC_INPUT_PREFIX = "TextToCodeValidateSubmissionV2/"
+TTC_METADATA_PREFIX = "TTCMetadata/"
+TTC_OUTPUT_PREFIX = "TTCOutput/"
+
+# AWS
+AWS_ACCESS_KEY_ID = "test_access_key_id"
+AWS_REGION = "us-east-1"
+AWS_SECRET_ACCESS_KEY = "test_secret_access_key"
+S3_BUCKET = "dibbs-text-to-code"
```
nickclyde: This pins `S3_BUCKET="dibbs-text-to-code"`, but PR #438 removes the `S3_BUCKET` env var entirely in favor of deriving it from the event payload. Once #438 lands, this line will be misleading (nobody reads it) and `e2e/test_e2e.py` references to `S3_BUCKET` will need to be plain constants. Probably worth dropping the variable here and the imports in the test file now, or at least flagging it in a TODO so whoever resolves the conflict knows which direction to go.

packages/data-curation/tests/test_augmentation.py
```diff
@@ -121,7 +123,7 @@ def test_build_augmented_loinc_files(self, cleanup_tmp_files):
         working_dir = pathlib.Path.cwd()
         if working_dir.name == "unit":
             input_path = pathlib.Path("assets") / "loinc_lab_names_20250930.csv"
-        elif working_dir.name == "dibbs-text-to-code":
+        elif working_dir.name == S3_BUCKET:
```
nickclyde: I don't think this comparison is semantically meaningful — `working_dir.name` is a directory name on the filesystem and `S3_BUCKET` is the S3 bucket name from `.env.test`. They only happen to both be the string `dibbs-text-to-code` today. If we ever change the bucket (or run tests inside a checkout with a different folder name), this branch will silently fall through to `RuntimeError`. Could we leave the literal `"dibbs-text-to-code"` here and add a comment if any clarification is needed?

packages/api/src/api/config.py
```diff
@@ -1,3 +1,4 @@
+from utils import get_env_var
 import os
```
nickclyde: Two nits:
1. `from utils import get_env_var` is a first-party import but sits above `import os` (stdlib) — isort will re-order this.
2. `import os` is no longer referenced in this file after `_get_env_variable` was removed; can it be removed?

e2e/handler.py
```diff
@@ -0,0 +1,44 @@
+"""Lambda handler: polls SQS (via event source mapping) for S3 event notificationsrouted through EventBridge, then reads the uploaded file from S3."""
```
nickclyde: I don't see anything that references this module from the new e2e tests — is it still needed? If it is, there's a typo in the docstring: `notificationsrouted` should be `notifications routed` (missing space).

packages/augmentation/src/augmentation/services/eicr_augmenter.py
```diff
@@ -198,9 +197,9 @@ def _get_new_effective_time(self) -> Element:
     def _get_new_version_number(self) -> Element:
         """Generate a versionNumber element for the augmented eICR document."""
         version_number_tag = etree.Element("versionNumber")
-        # hard code to 1 for now
-        # TODO: we may need to have some way to increment this later
-        version_number_tag.set("value", "1")
+        old_version_number = self._get_old_version_number().get("value")
+        new_version_number = int(old_version_number) + 1 if old_version_number is not None else 1
+        version_number_tag.set("value", str(new_version_number))
```
nickclyde: This is a meaningful behavior change — previously we hard-coded `versionNumber=1` (with a TODO noting this might need to be incremented), and now we're doing `old + 1`. Just wanted to call it out for the rest of the team to review and make sure we agree this is how we want to do it. Also, probably worth adding an explanatory comment.

packages/text-to-code-lambda/tests/test_lambda_function.py
```diff
@@ -162,46 +76,6 @@ def test_handler_with_empty_body(self, example_sqs_event, caplog_warning, mock_o
         }
```
nickclyde, line 162: A bunch of tests got removed in this file: `test_handler_saves_metadata_when_no_relevant_schematron_fields`, `test_handler_continues_when_selected_candidate_is_none`, `test_handler_adds_unmatched_error_when_selected_candidate_has_no_opensearch_hits`, `test_process_record_pipeline_returns_no_matches_found_when_no_candidates_are_selected`. These cover real error paths that still exist in the new code (no schematron fields, no candidate selected, empty OpenSearch results). I don't think the snapshot-based happy-path test doesn't replace them. What was your rationale for removing them?

packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
```diff
@@ -1,313 +1,48 @@
...
+    # Assert that the augmentated eICR  was saved to S3
+    augmentation_metadata = lambda_handler.get_file_content_from_s3(
+        bucket_name=S3_BUCKET,
+        object_key=f"{AUGMENTATION_METADATA_PREFIX}{mock_aws_setup.persistence_id}",
+    )
 
-        assert result["statusCode"] == SUCCESS_CODE
-        assert result["message"] == "Augmentation processed successfully!"
-        assert result["num_success_eicrs"] == 1
+    snapshot.assert_match(augmentation_metadata, "augmentation_metadata.json")
```
nickclyde: Several tests were removed here as well — `TestParseNonstandardCodes` (helper is gone, that's fair), `test_handler_source_bucket_routing` (actively tests the PR #438 behavior we want to preserve), `test_handler_error_missing_eicr`, `test_handler_error_missing_ttc_output`, `test_handler_mixed_batch_results`, `test_handler_skips_empty_sqs_body`. What's the reasoning for removing these?

pyproject.toml
JNygaard-Skylight: [`[tool.pytest]` was added in pytest v9](https://docs.pytest.org/en/stable/reference/customize.html#pyproject-toml).  so it should work for all of us as we are pinned to `"pytest>=9.0.3"`.  It's a little bit cleaner as it takes advantage of TOML types, but I think for our current configuration it does not appear to make a difference. I'll revert it to minimize changes.

e2e/test_e2e.py
```diff
@@ -0,0 +1,280 @@
...
+
+from augmentation_lambda.lambda_function import handler as augmentation_lambda
+from text_to_code_lambda.lambda_function import handler as ttc_handler
+
+AUGMENTATION_METADATA_PREFIX = "AugmentationMetadataV2/"
+AUGMENTED_EICR_PREFIX = "AugmentationEICRV2/"
+AWS_ACCESS_KEY_ID = "test_access_key_id"
+AWS_SECRET_ACCESS_KEY = "test_secret_access_key"  # noqa: S105
+EICR_INPUT_PREFIX = "eCRMessageV2/"
+REGION = "us-east-1"
```
robertandremitchell, line 22: very minor, but given that our local instance of these folders are in us-east-2, could be good to keep that and `.env.test` in alignment

robertandremitchell (approved):
looks awesome! some small non-blocking nits: the region change i called out in the comment, and also there seem to be some weird whitespace-only updates to some files. they don't trigger any ruff issues when I run ruff locally, so it's probably fine.

PR #438, opened 2026-04-07
feat: derive S3 bucket from event instead of static env var

geo-ruvos:
## Description

Switches the TTC Lambda from using a static `S3_BUCKET` environment variable to deriving the bucket name directly from the incoming S3 event payload (`detail.bucket.name`). This source-bucket directed architecture allows a single deployed Lambda to serve multiple data pipelines — in the case of AIMS, both the eCR pipeline (`ecr-data-repository`) and the TTC training pipeline (`ecr-ttc-training`) — without any code or configuration changes.

Key changes:
- Removed the `S3_BUCKET` env var; bucket is now required from the event and a `ValueError` is raised if missing.
- Added structured logging of `event_bucket` and `event_object_key` at the start of each record, with `bucket_name` carried as a logger context key through the full pipeline for easier CloudWatch debugging.
- Added a test covering the missing-bucket error case.
- Wrote README documentation covering the source-bucket architecture, IAM requirements (read/write on all event-producing buckets), and environment variables.

## Related Issues

N/A

## Additional Notes

The Lambda's IAM execution role must be updated to grant `s3:GetObject` and `s3:PutObject` on any bucket that will produce events for this Lambda. When onboarding a new pipeline bucket, the IAM policy must be extended accordingly.

## Checklist

- [x] I have ensured that the pull request is of a manageable size, allowing it to be reviewed within a single session.
- [x] I have reviewed my changes to ensure they are clear, concise, and well-documented.
- [x] I have updated the documentation, if applicable.
- [x] I have added or updated test cases to cover my changes, if applicable.

## Checklist for Reviewers


packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -106,16 +105,30 @@ def process_record(record: SQSRecord, s3_client: BaseClient, opensearch_client:
...
-    logger.info(f"Processing S3 Object: s3://{bucket_name}/{object_key}")
+    bucket_name = eventbridge_data.get("bucket_name")
+
+    if not bucket_name:
+        raise ValueError(
+            "No bucket name found in S3 event payload. "
+            "The TTC lambda derives its target bucket from the event and does not use a "
+            "static bucket configuration. Ensure the EventBridge/S3 event includes "
+            "detail.bucket.name."
+        )
```
nickclyde, line 115: I think this `ValueError` is unreachable. `get_eventbridge_data_from_s3_event` in `lambda_handler/lambda_handler.py:135` does a direct key access:

```python
bucket_name = event["detail"]["bucket"]["name"]
```

If `detail.bucket.name` is missing from the event payload, a `KeyError` is raised inside `get_eventbridge_data_from_s3_event` *before* control returns here. The `.get("bucket_name")` on line 108 operates on the *returned dict*, which always has the key set if the function returns successfully.

To make this guard work as intended, you'd need to either:
1. Change `get_eventbridge_data_from_s3_event` to use `.get()` so it returns `None` instead of raising, or
2. Catch the `KeyError` here and re-raise as the `ValueError` with the descriptive message.

Option 1 is probably cleaner since it keeps the error handling in the consumer, but it touches the shared `lambda_handler` package.

packages/text-to-code-lambda/tests/test_lambda_function.py
```diff
@@ -162,6 +162,18 @@ def test_handler_with_empty_body(self, example_sqs_event, caplog_warning, mock_o
...
+    def test_handler_fails_when_event_has_no_bucket(self, example_sqs_event, mock_opensearch):
+        """Test handler reports failure when S3 event payload is missing a bucket name."""
+        payload = json.loads(example_sqs_event["Records"][0]["body"])
+        del payload["detail"]["bucket"]["name"]
+        example_sqs_event["Records"][0]["body"] = json.dumps(payload)
+
+        resp = lambda_function.handler(example_sqs_event, {})
+
+        assert resp["num_failure_eicrs"] == 1
+        assert resp["num_success_eicrs"] == 0
```
nickclyde, line 174: Our repo settings block CI from running on external contributions, but when I run unit tests locally, this test fails as written due to the bug mentioned above. When `del payload["detail"]["bucket"]["name"]` removes the key, the call to `get_eventbridge_data_from_s3_event` raises `KeyError('name')`, not the `ValueError` from `process_record`. So `str(e)` in the handler's `except` block will be `"'name'"`, and the assertion `"No bucket name found" in resp["failures"][0]["error"]` will be `False`.

Once the issue is fixed, this test should work correctly.

nickclyde (commented):
Thanks for this, Geo! Removing the static `S3_BUCKET` fallback in favor of event-driven bucket resolution is a good simplification. However, I believe there's a small bug, and a few other notes. One minor thing here since it's outside the files you touched: `terraform/main.tf:428` still injects `S3_BUCKET` as an env var for the TTC Lambda. It's now dead config and should be removed (along with the variable declaration) to avoid confusion.

packages/text-to-code-lambda/README.md
```diff
@@ -0,0 +1,37 @@
...
+| Variable | Required | Default | Description |
+|---|---|---|---|
+| `EICR_INPUT_PREFIX` | No | `eCRMessageV2/` | S3 key prefix for original eICR documents |
+| `SCHEMATRON_ERROR_PREFIX` | No | `ValidationResponseV2/` | S3 key prefix for schematron validation responses |
+| `TTC_INPUT_PREFIX` | No | `TextToCodeSubmissionV2/` | S3 key prefix for TTC submission triggers |
+| `TTC_OUTPUT_PREFIX` | No | `TTCAugmentationMetadataV2/` | S3 key prefix for TTC augmentation output |
+| `TTC_METADATA_PREFIX` | No | `TTCMetadataV2/` | S3 key prefix for TTC analysis metadata |
+| `AWS_REGION` | Yes | — | AWS region |
+| `OPENSEARCH_ENDPOINT_URL` | Yes | — | OpenSearch cluster endpoint |
+| `OPENSEARCH_INDEX` | No | `ttc-index` | OpenSearch index name for vector search |
```
nickclyde, line 36: Nit: The env var table lists `AWS_REGION` as required, but the Lambda runtime sets this automatically (it's injected by the Lambda execution environment). Might be worth marking it as auto-provided rather than required to avoid confusion for someone trying to configure this locally.

geo-ruvos:
@nickclyde I think I have addressed your notes. Let me know how it looks now!

nickclyde (changes requested):
Hi Geo - the unit tests are passing for me now! One more small change before we can merge: with PR #465, we removed all references to `EICR_INPUT_PREFIX`, so if you could remove the references in your changeset, we should be good to go.

packages/text-to-code-lambda/tests/conftest.py
```diff
@@ -16,6 +16,7 @@
 from lambda_handler.test_constants import TEST_PERSISTENCE_ID
 from text_to_code_lambda import lambda_function
 
+EICR_INPUT_PREFIX = "eCRMessageV2/"
```
nickclyde: Sorry to be nitpicky 😅 but there's another `EICR_INPUT_PREFIX` here

packages/text-to-code-lambda/tests/conftest.py
```diff
@@ -25,7 +26,7 @@
 
 def pytest_configure() -> None:
     """Configure env variables for pytest."""
-    os.environ["S3_BUCKET"] = S3_BUCKET
+    os.environ["EICR_INPUT_PREFIX"] = EICR_INPUT_PREFIX
```
nickclyde: Another reference

packages/text-to-code-lambda/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Logging
+
+Every TTC invocation logs the **event bucket** and **object key** as structured fields at the start of record processing, and carries the **bucket name** and **persistence ID** as context keys through all subsequent log lines. This makes it straightforward to filter CloudWatch logs by bucket or object when debugging cross-pipeline issues.
+
+## Environment Variables
+
+| Variable | Required | Default | Description |
+|---|---|---|---|
+| `EICR_INPUT_PREFIX` | No | `eCRMessageV2/` | S3 key prefix for original eICR documents |
```
nickclyde: Another reference

nickclyde (changes requested):
Sorry for the nitpicking, just want to avoid any confusion!

geo-ruvos:
> Sorry for the nitpicking, just want to avoid any confusion!

No worries at all... my mistake, I was not thorough enough!  I will fix this now.

nickclyde (approved):

PR #446, opened 2026-04-08
Add index push to github workflow

bamader:
## Description

This PR adds a push to APHL's eCR Repo for the Index Lambda docker container on merge-to-main for our repo. Previously, only TTC and Augmentation were pushed; now, the workflow file has been modified to also push the Index Lambda whenever changes are detected to its file contents (using the same check which we use to build and push the Index image to GHCR).

## Related Issues

Closes #436 

## Additional Notes

I _think_ we'll be okay without a separate repo URL, since on line 109 we push the image to a location based on its name (TTC, Augmentation, or Index), but if that's wrong I can definitely add another variable.

### DIFF
--- a/.github/workflows/docker-image-push.yml
+++ b/.github/workflows/docker-image-push.yml
@@ -74,7 +74,6 @@ jobs:
             type=raw,value=latest,enable=${{ github.ref_name == 'main' }}
 
       - name: Build and push image
-        if: ${{ matrix.name != 'index' || github.event_name == 'workflow_dispatch' || steps.changes.outputs.index == 'true' }}
         uses: docker/build-push-action@v7
         with:
           context: .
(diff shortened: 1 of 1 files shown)


robertandremitchell (approved):

nickclyde (changes requested):
I think we can actually take out the if checks on second thought! Because they are currently just excluding index from the push.

One other thing - I think I need to ask APHL to set up a separate ECR repo for our index image!

BradySkylight (approved):
Looks simple enough.  Nice!

bamader:
@nickclyde Sounds good, I'll remove the `if` checks altogether then, and we can just do wholesale builds and pushes for each image! 

For the new index lambda repo, that sounds good too. I'll leave this open until you hear back from APHL on setting up that extra repo, since it looks like we'll need to set up a GitHub repo secret to hold that information, and I don't want to cause any errors by trying to push with an illegal value or something like that.

nickclyde:
@bamader I just heard from Sean that they have the `index` repo set up now!

nickclyde (approved):

PR #448, opened 2026-04-08
Fix Lambda-incompatible Docker image manifest in dev push workflow

nickclyde:
## Summary
- APHL reported that Lambda still rejects TTC container images with `InvalidParameterValueException` because the image manifest uses OCI format instead of Docker v2
- #428 Added `provenance: false` to the `docker/build-push-action` step to produce Docker v2 manifests (matching the fix already in `release.yml`)
- This PR replaces `docker buildx imagetools create` with `docker pull`/`docker tag`/`docker push` for the APHL ECR push step — `imagetools create` can convert the manifest format when copying between registries, negating the `provenance: false` fix

Reference: CDCgov/dibbs-ecr-refiner hit the same issue and applied a similar fix.

## Test plan
- [ ] Trigger the `docker-image-push.yml` workflow via manual dispatch
- [ ] Verify the APHL ECR image manifest format: should be `application/vnd.docker.distribution.manifest.v2+json`
- [ ] Confirm APHL can successfully create Lambda functions from the new images


BradySkylight (approved):
Nice catch and easy fix!  🚀

PR #457, opened 2026-04-09
Add clear_index action to index Lambda

nickclyde:
## Summary
- Adds a `clear_index` action to the index Lambda that deletes and recreates the OpenSearch index, allowing a clean slate before OSIS re-ingestion
- The OSIS ingestion pipeline creates new documents with auto-generated IDs on each run, so without clearing first, re-ingesting updated embeddings results in duplicates
- Default `create_index` behavior is unchanged — no impact on existing Terraform invocations

## Usage
Before triggering the OSIS pipeline with new embeddings:
```bash
aws lambda invoke \
  --function-name ttc-index-lambda \
  --payload '{"action": "clear_index"}' \
  /tmp/response.json
```

## Test plan
- [x] Existing index-lambda tests pass
- [x] New tests for `clear_index` action (with and without existing index)
- [x] Ruff linting passes
- [ ] Deploy to dev and verify `clear_index` deletes/recreates the index before re-ingestion


BradySkylight (approved):
This looks good and makes sense.  The only question I have is how often is this run?  Will this just be run when updating embeddings?  Hopefully not everytime we spin up Opensearch or the IndexLambda.

packages/index-lambda/src/index_lambda/lambda_function.py
```diff
@@ -1,70 +1,93 @@
...
+
+    if action == "clear_index":
+        return _clear_index(os_client, index_name)
+    return _create_index(os_client, index_name)
 
-    # Check that index exists
-    status = False
+
+def _clear_index(os_client: OpenSearch, index_name: str) -> dict:
+    """Delete the index if it exists, then recreate it with correct mappings."""
```
robertandremitchell: non-blocking, but should probably add :param: for each of these new functions?

robertandremitchell (approved):
this makes sense to me. I'm of two minds as to whether we want these OpenSearch functions siloed to `lambda-handler`, sort of how a lot of the AWS generic functions are. I don't think it's necessary, since we shouldn't need create/clear indices across other lambdas.

PR #459, opened 2026-04-13
Actually deploy on merge to main

nickclyde:
This will make it so we actually deploy to AWS on merge to main. Also, removes the AWS_REGION env var from the augmentation lambda since that gets read automatically by lambdas.


BradySkylight (approved):
🚀

bamader (approved):

PR #465, opened 2026-04-13
Consolidate eICR input prefixes into TTC_INPUT_PREFIX

nickclyde:
## Summary
- Drop the `EICR_INPUT_PREFIX` env var. Both the TTC and Augmentation Lambdas now read the original eICR from `TTC_INPUT_PREFIX` (default `TextToCodeSubmissionV2/`), which is also the EventBridge trigger prefix.
- Update Terraform IAM read policies, Lambda env vars, and the `eicr_input_prefix` variable accordingly; refresh `CLAUDE.md` and `terraform/README.md`.

This fixes the deployed-pipeline mismatch where the S3 event triggered on `TextToCodeSubmissionV2/` but the Lambda tried to fetch the eICR from `eCRMessageV2/`.

## Test plan
- [x] `uv run pytest packages/text-to-code-lambda packages/augmentation-lambda`
- [x] `uv run ruff check`
- [x] `terraform fmt -check && terraform validate`
- [ ] End-to-end: upload an eICR to `s3://<bucket>/TextToCodeSubmissionV2/<persistence_id>` in the deployed environment and confirm both Lambdas succeed

### DIFF
--- a/packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
+++ b/packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
@@ -20,7 +20,7 @@ logger = structlog.get_logger()
 
 # Environment variables
 S3_BUCKET = os.getenv("S3_BUCKET", "dibbs-text-to-code")
-EICR_INPUT_PREFIX = os.getenv("EICR_INPUT_PREFIX", "eCRMessageV2/")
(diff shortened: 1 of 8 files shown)


BradySkylight (approved):
Makes it easier to read and understanding what the prefix buckets are for.  Nice!

robertandremitchell (approved):

Issue #484, opened 2026-04-21
Centralize test environment variables


JNygaard-Skylight:
Instead of setting enviroment variables manually or relying on hardcoded defaults, with the [pytest-env](https://pypi.org/project/pytest-env/) plugin all environment variables used by tests can be set in a single file, `.env.test`.

PR #491, opened 2026-04-22
refactor: Required environment variable logic

### LINKED ISSUE #483: Deduplicate environment variable logic
Two different functions for getting a required environment variable are defined in the repo:
- `_get_env_variable` in `packages/api/src/api/config.py`
- `require_env` in `packages/lambda-handler/src/lambda_handler/lambda_handler.py`

A single function, defined in `packages/utils` could be used instead.

JNygaard-Skylight:
## Description

Removes duplicated logic for getting requred environment variables and replaces them with a single function in the utils package.

## Related Issues

Closes #483


packages/utils/src/utils/utils.py
```diff
@@ -0,0 +1,19 @@
+from os import getenv
+
+
+def get_env_variable(name: str) -> str:
+    """Grabs a variable by name from the environment. Throws an error if the variable is not present.
```
robertandremitchell, line 5: doc block doesn't match our standard one with :param:.

robertandremitchell (commented):
looks great! do we want to add this to `augmentation-lambda` and `text-to-code-lambda`'s respective `lambda_function.py` as well? right now that's just using the os library to get env vars.

JNygaard-Skylight:
> looks great! do we want to add this to `augmentation-lambda` and `text-to-code-lambda`'s respective `lambda_function.py` as well? right now that's just using the os library to get env vars.

I think so, but I'm still not quite sure how we should handle environment variables, apart from removing the duplicate logic, so I'm leaving that for different ticket #484

BradySkylight (approved):
Nice!  I like having this function in utils vs. in a model

PR #499, opened 2026-04-23
refactor: Declassify embedder and reranker

JNygaard-Skylight:
## Description

Currently the embedder and reranker methods are in wrapper classes. There is no need for them to be in a class. Embedder has no state at all, and the reranker does have state but at the module level (which I suppose is a code smell, but seems reasonable for such a simple module in this situation).

Also adds a `ScoredResult` class for better typing of the `rerank` function.

## Related Issues

Closes # n/a

## Additional Notes
If anyone has  a reason to keep these classes, I'm open to keep them, but this seemed like a pretty easy little refactor that simplifies things a little with no downside.

### DIFF
--- a/packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
+++ b/packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
@@ -19,11 +19,11 @@ from text_to_code.models import Candidate
 from text_to_code.models import SchematronErrorDetail
 from text_to_code.models import query as query_models
 from text_to_code.services import eicr_processor
-from text_to_code.services import embedder
 from text_to_code.services import evaluator
-from text_to_code.services import reranker
 from text_to_code.services import schematron_processor
+from text_to_code.services.embedder import embed
 from text_to_code.services.query import QueryBuilder
+from text_to_code.services.reranker import rerank
 
 # Initialize the logger
 logger = Logger(service="ttc")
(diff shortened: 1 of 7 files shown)


bamader (commented):
Agree with removing classes for the Embedder and Reranker, but just my two cents, I would personally opt not to make a `ScoredResult` class. Candidly, I think our repo already has too many classes, to the point of it feeling burdensome and non-pythonic, but it's not a hill I'll die on. Curious what other devs think too, and happy to go with whatever the consensus is.

BradySkylight:
I like the simplification for sure.  If we can make all the code in this repo more python-like, great!  If it's too much work for just re-factoring sake then we can avoid that....though I would still keep these changes.

JNygaard-Skylight:
> Agree with removing classes for the Embedder and Reranker, but just my two cents, I would personally opt not to make a `ScoredResult` class. Candidly, I think our repo already has too many classes, to the point of it feeling burdensome and non-pythonic, but it's not a hill I'll die on. Curious what other devs think too, and happy to go with whatever the consensus is.

Also curious what other devs think but my thinking is that a type hint with just `dict` is no better than no typing hinting at all. Ty will see the output of `rerank` as `list[dict[Unknown, Unknown]]` which is not very helpfully and will not prevent any errors.

Of course there are situations when we will generally not know the structure of a dictionary, but in this case we know exactly what the structure is so there is no reason not to make that visible to the type checker. 

Also, this is not really a class, as the output of `rerank` is still just a dictionary. This is just saying that the dictionary will have the structure of the class `ScoredResult`. 

I would say that I have been a little too enthusiastic implementing things as Pydantic models, when a simple `@dataClass` or `TypedDict` would work, as we are not usually taking advantage of Pydantic's features.

bamader (approved):
Great call-out that this is just a simple TypedDict and not a full pydantic model. I definitely missed that on first read due to my own class fatigue, but that's on me, not the code. I agree with the reasoning that this is a very simple data holder that provides some internal clarity in the code without convoluting any extra validation or BaseModels.

These changes are good so switching to approval, but it sounds like in our tech debt squashing window next week, we might benefit from an eng discussion around possible class simplifications and increasing "pythonicity" (it's a word don't @ me). Even if no change comes out immediately, honestly it could be good to just get all the engineers on the same page and iron out a philosophy. That's a conversation for another time though, so +1'ing this clean up 💪

PR #500, opened 2026-04-23
Using snapshots for augementation and ttc lambdas

### LINKED ISSUE #485: Snapshots can be used for Lambda function tests
Currently the lambda function tests, `packages/text-to-code-lambda/tests/test_lambda_function.py` and `packages/augmentation-lambda/tests/test_augmentation_lambda_function.py` individually assert that each field matches the expected value. 

Instead, this would be a good use of a snapshot test. This would significantly reduce the amount of boilerplate, make it easier to update, and would more likely catch unexpected changes.

robertandremitchell:
## Description

Using snapshots for some of the ttc and augmentation lambda tests to automate/ease burden of tests, more in line with e2e
## Related Issues

Closes #485 

## Additional Notes

## Checklist

## Checklist for Reviewers

### DIFF
--- a/packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
+++ b/packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from uuid import UUID
 from zoneinfo import ZoneInfo
 
+import pytest
 import time_machine
 from pytest_mock import MockerFixture
 from pytest_snapshot.plugin import Snapshot
@@ -95,7 +96,8 @@ def _build_empty_body_event() -> dict[str, object]:
     }
 
 
-def _mock_augmented_ids(mocker: MockerFixture) -> None:
+@pytest.fixture(autouse=False)
+def mock_augmented_ids(mocker: MockerFixture) -> None:
     doc_id = UUID("12345678-1234-5678-1234-567812345678")
     set_id = UUID("87654321-4321-8765-4321-876543218765")
     mocker.patch("augmentation.services.eicr_augmenter.uuid4", side_effect=[doc_id, set_id])
@@ -151,9 +153,13 @@ class TestHandler:
         assert result["num_success_eicrs"] == 1
 
     def test_handler_writes_outputs_to_s3(
-        self, example_sqs_event, mock_aws_setup, mocker: MockerFixture, snapshot: Snapshot
+        self,
+        example_sqs_event,
+        mock_aws_setup,
+        mocker: MockerFixture,
+        snapshot: Snapshot,
+        mock_augmented_ids,
     ) -> None:
-        _mock_augmented_ids(mocker)
 
         with time_machine.travel(
             datetime(2026, 2, 13, 15, 27, 57, tzinfo=ZoneInfo("America/New_York")), tick=False
@@ -186,6 +192,7 @@ class TestHandler:
         mock_aws_setup,
         mocker: MockerFixture,
         snapshot: Snapshot,
+        mock_augmented_ids,
     ) -> None:
         """Verify bucket name is extracted from the S3 event, not the env var."""
         custom_bucket = "custom-bucket"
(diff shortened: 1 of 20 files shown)


bamader (commented):
All of the logic here looks fine, but I have to ask: is this a desirable change, switching to snapshot tests? This modifies things in only 2 files yet introduces 18 new files, which feels like a pretty big "sea of files" bloat. Imo this also makes it harder to check what's being tested because tests aren't contained to one place anymore.

Just my opinion, so would love to hear from others on this, and I'm fine going whatever direction we decide; just raising the point because I don't think we have any established "norms" so to speak on repo structure and best practices, but in the wake of considering tech debt refactors it feels like maybe we should iron those out so we can actually tell what's deliberate and what's debt.

BradySkylight:
Might be something to discuss?  I am not sure exactly what is happening here and don't feel like I can give feedback for the affirmative or otherwise at this point.

packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
```diff
@@ -36,6 +43,64 @@
...
+            _build_sqs_record(
+                body="",
+                message_id="msg-empty-body",
+                receipt_handle="test-receipt-handle",
+            )
+        ]
+    }
+
+
+def _mock_augmented_ids(mocker: MockerFixture) -> None:
```
JNygaard-Skylight: I have no idea if this is actually any better (perhaps more idiomatic, idk) but this could be written as a Pytest fixture:
```python
@pytest.fixture(autouse=False)
def mock_augmented_ids(mocker: MockerFixture) -> None:
    doc_id = UUID("12345678-1234-5678-1234-567812345678")
    set_id = UUID("87654321-4321-8765-4321-876543218765")
    mocker.patch("augmentation.services.eicr_augmenter.uuid4", side_effect=[doc_id, set_id])
```
then:
```python
    def test_handler_writes_outputs_to_s3(
        self,
        example_sqs_event,
        mock_aws_setup,
        snapshot: Snapshot,
        mock_augmented_ids,
    ) -> None:
# ...
    def test_handler_source_bucket_routing(
        self,
        example_s3_event_payload,
        mock_aws_setup,
        snapshot: Snapshot,
        mock_augmented_ids,
    ) -> None:
```

What do you think?

robertandremitchell:
> All of the logic here looks fine, but I have to ask: is this a desirable change, switching to snapshot tests? This modifies things in only 2 files yet introduces 18 new files, which feels like a pretty big "sea of files" bloat. Imo this also makes it harder to check what's being tested because tests aren't contained to one place anymore.
> 
> Just my opinion, so would love to hear from others on this, and I'm fine going whatever direction we decide; just raising the point because I don't think we have any established "norms" so to speak on repo structure and best practices, but in the wake of considering tech debt refactors it feels like maybe we should iron those out so we can actually tell what's deliberate and what's debt.

I'm of two minds. In prior work, I've found snapshots to be helpful in flagging breaking changes. They're also sometimes a bit annoying to update, and definitely get the concern that abstracting code into separate folders can make it harder to see/understand what exactly is being tested.

JNygaard-Skylight:
> All of the logic here looks fine, but I have to ask: is this a desirable change, switching to snapshot tests? This modifies things in only 2 files yet introduces 18 new files, which feels like a pretty big "sea of files" bloat. Imo this also makes it harder to check what's being tested because tests aren't contained to one place anymore.
> 
> Just my opinion, so would love to hear from others on this, and I'm fine going whatever direction we decide; just raising the point because I don't think we have any established "norms" so to speak on repo structure and best practices, but in the wake of considering tech debt refactors it feels like maybe we should iron those out so we can actually tell what's deliberate and what's debt.

I think snapshots are a good fit for testing the output of the lambda handlers because otherwise we need to individually assert the value of each field, like what is currently happening in `packages/text-to-code-lambda/tests/test_lambda_function.py`. Snapshots will also catch when new fields are added, and are easier to update.

I would not worry about file bloat. These are not code files and are not meant to be manually edited or even looked at unless working on the test in question.

packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
robertandremitchell: makes sense to me. updated!

JNygaard-Skylight (approved):
📸

PR #503, opened 2026-04-24
test: experimenting with benchmarking

JNygaard-Skylight:
## Description

Now that we are going to be moving into refactoring work and because of the performance-sensitive nature of TTC we will want a way to track performance to spot regressions and to see if changes to improve performance indeed help. 

I've so far only created a single benchmark of the the reranker and embedder because they are the two slowest parts of our entire system.

I m using [pytest-benchmark](https://pytest-benchmark.readthedocs.io/en/latest/) for the benchmarks, and [Bencher](https://bencher.dev/) to keep track of performance history.

To be completely honest, I am no expert in benchmarking and have never used Bencher before so I'm not entirely sure how best to configure it and mostly copied [the examples from Bencher's docs](https://bencher.dev/docs/how-to/github-actions). so this will be a learning experience. But hopefully this PR will create a baseline benchmark that future PRs can be compared to.

## Related Issues

Closes # n/a


robertandremitchell (commented):
makes sense to me. could be worth adding a note to the README about running benchmarks.

only reason I'm not approving is because I have basically no prior experience with benchmarking, but the out-the-box solution from pytest seems sufficient to me fwiw.

JNygaard-Skylight:
> makes sense to me. could be worth adding a note to the README about running benchmarks.
> 
> only reason I'm not approving is because I have basically no prior experience with benchmarking, but the out-the-box solution from pytest seems sufficient to me fwiw.

Tbh, this PR is more about getting the conversation started. My thinking is that we need a more robust system than just using `pytest --durations` because, at least for me, the test durations are so inconsistent, it's hard to tell if you've made a difference, or in what direction. And that's not even trying to compare results on different machines. 

Idk, I think this is something we'll want to keep in mind, especially with my understanding of the scale of the eICRs APHL handles.

PR #509, opened 2026-04-28
Standardize logging across packages

### LINKED ISSUE #506: Standardize logging across packages
**Summary**
The codebase currently uses three different logging mechanisms across closely-related Lambda code: `structlog`, `aws_lambda_powertools.Logger`, and stdlib `logging`.

**Why**
Inconsistent log shapes (key-value vs. printf, different correlation-ID conventions) make CloudWatch aggregation harder and hide the relationship between a TTC invocation and the augmentation job that follows it. Pick one and keep it.

**Scope**
- Decide on the standard (recommend `aws_lambda_powertools.Logger` because the TTC Lambda already uses it and Powertools provides correlation-ID injection from SQS/S3 events for free).
- Migrate `packages/augmentation-lambda` and any text-to-code services to it.
- Add a section to CLAUDE.md and `terraform/README.md` documenting the chosen pattern.

**Acceptance criteria**
- One logging framework used in all `packages/*-lambda` and the core library packages.
- Lambda invocations emit structured JSON logs with consistent keys (request ID, S3 key, status).

**References**
- `packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py:5,19` (structlog)
- `packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py:7,29` (powertools)
- `packages/text-to-code/src/text_to_code/services/schematron_processor.py:1,10` (stdlib)
- `packages/text-to-code/src/text_to_code/services/eicr_processor.py:66` — broad `except Exception`

robertandremitchell:
## Description

Switching and standardizing logging to use aws powertools

Some q's:

1. Do we want this logging on `index-lambda` and/or `lambda-handler`?
2. Are there any other parts of the `text-to-code` and `augmentation` base functions we import we want to add logging similar to eicr_processor.py and schematron_processor.py?
3. Are we good with the structure of these logs? Should more or less info be added?

## Related Issues

Closes #506 

## Additional Notes

## Checklist

## Checklist for Reviewers


packages/lambda-handler/pyproject.toml
```diff
@@ -3,6 +3,7 @@ name = "lambda-handler"
 version = "0.1.0"
 readme = "README.md"
 dependencies = [
+  "aws-lambda-powertools>=2.0.0",
```
nickclyde, line 6: Are we using`aws_lambda_powertools` anywhere under `packages/lambda-handler/src/lambda_handler/` yet? I think this dep can be removed unless we're planning to add Powertools logging to the helper functions in `lambda_handler.py` (which might actually be a good idea, since `get_file_content_from_s3`, `put_file`, etc. are good places for logs). What do you think?

packages/text-to-code/pyproject.toml
```diff
@@ -12,6 +12,7 @@ dependencies = [
   # to ensure compatibility across all hardware.
   # Once Intel Mac support is no longer needed, we can remove these pinned torch and numpy
   # versions and rely on spacy and sentence-transformers to manage their own dependencies.
+  "aws-lambda-powertools>=2.0.0",
```
nickclyde: Adding `aws-lambda-powertools` as a dep on `text-to-code` brings a few things to mind. The `text-to-code` package is also imported by non-Lambda callers in this repo (e.g. data-curation, evaluator scripts). Powertools works fine outside Lambda, but I think its default log level is WARNING when no parent Logger is configured, so the child loggers in `eicr_processor`/`schematron_processor` will be silent for non-Lambda callers unless they set `POWERTOOLS_LOG_LEVEL`. Should we document that in the README, or alternatively, keep `text-to-code` on stdlib `logging` and let the Lambda wrappers configure the root logger? Or do we not really care for data-curation and evaluator?

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -49,10 +50,15 @@ def handler(event: SQSEvent, context: LambdaContext) -> dict:
             _process_record(record, s3_client)
             successes.append(record.message_id)
         except Exception as e:
-            logger.exception(f"Error processing record: {e}", message_id=record.message_id)
+            logger.exception(
+                "Error processing record",
+                message_id=record.message_id,
+                error=str(e),
+                status="error",
+            )
```
nickclyde, line 57: I believe `error=str(e)` is redundant here. Powertools `logger.exception(...)` automatically captures the exception message and traceback under `exception` / `exception_name` keys, so I think adding `error=str(e)` will the same string twice in the log payload. I think we should be able to drop `error=str(e)` and let Powertools handle it. Same comment applies to the equivalent block in the TTC lambda (around L70-75).

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -76,46 +91,53 @@ def _process_record(record: SQSRecord, s3_client: BaseClient) -> None:
...
-    ttc_output = _load_ttc_output(persistence_id, s3_client, bucket_name)
-    original_eicr = _load_original_eicr(persistence_id, s3_client, bucket_name)
-    nonstandard_codes = _parse_nonstandard_codes(ttc_output)
+    with logger.append_context_keys(
+        persistence_id=persistence_id,
+        bucket_name=bucket_name,
+        s3_key=object_key,
+    ):
+        logger.info("Processing S3 object", status="processing")
+        logger.info("Extracted persistence_id", status="processing")
```
nickclyde: Two thoughts:

1. We set `s3_key=object_key` in the context, but every nested call (`_load_ttc_output`, `_load_original_eicr`, `_save_augmentation_outputs`) also sets its own `s3_key=` with a different value, which silently overrides the context's `s3_key` for that single log line. I think the trigger key in the context ends up being misleading, since most logs in this block won't carry it. Could we rename it to `trigger_s3_key` / `event_s3_key`) so we can tell which key any given log line refers to?

2. `logger.info("Extracted persistence_id", status="processing")` on L111 doesn't carry the value in the message anymore (it's only in context). And it fires immediately after `"Processing S3 object"` with the same status. I think we can collapse these two into a single log line, or drop the second one entirely.

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -121,8 +138,9 @@ def process_record(record: SQSRecord, s3_client: BaseClient, opensearch_client:
     with logger.append_context_keys(
         persistence_id=persistence_id,
         bucket_name=bucket_name,
+        s3_key=object_key,
     ):
```
nickclyde, line 134: Same comment as in augmentation-lambda: `s3_key=object_key` in the context gets overridden by every helper that passes its own `s3_key` (`_load_schematron_data_fields` writes the schematron key, `_load_original_eicr` writes the eICR key, `_save_ttc_outputs` writes the output keys, etc.). Could we drop it from the context, or rename to `trigger_s3_key` for clarity?

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -399,22 +452,35 @@ def _process_record_pipeline(
...
             bucket_name=bucket_name,
             object_key=f"{TTC_METADATA_PREFIX}{persistence_id.removesuffix('.xml')}.json",
             s3_client=s3_client,
         )
+        logger.info(
+            "Saved TTC metadata output to S3",
+            bucket_name=bucket_name,
+            s3_key=f"{TTC_METADATA_PREFIX}{persistence_id.removesuffix('.xml')}.json",
+            status="success",
+        )
```
nickclyde: I think this no-data-fields branch is now duplicating the metadata-save logic from `_save_ttc_outputs` (the `put_file` call plus the surrounding "Saving" / "Saved" log pair). Could we extract a small `_save_ttc_metadata_output(...)` helper and call it from both `_save_ttc_outputs` and this branch? It would keep log message wording in sync and avoid drift if we ever change the log keys.

packages/index-lambda/src/index_lambda/lambda_function.py
```diff
@@ -84,10 +84,6 @@ def _set_slowlog(os_client: OpenSearch, index_name: str, threshold_ms: int) -> d
     settings = os_client.indices.get_settings(index=index_name)
     mappings = os_client.indices.get_mapping(index=index_name)
```
nickclyde, line 109: Glad to see the bare `print()`s go. While we're here, would it be worth adding a Powertools `Logger(service="ttc-index")` (or similar) and at least logging the entry/exit of `handler` plus the chosen `action`? Right now we have no structured trace. Could be a follow-up if we want to keep the PR scoped.

packages/text-to-code-lambda/tests/test_lambda_function.py
```diff
@@ -44,6 +44,13 @@ def _serialize_ttc_metadata_snapshot(ttc_metadata_output: dict[str, object]) ->
     return json.dumps(normalized, indent=2, sort_keys=True)
 
 
+class MockLambdaContext:
+    function_name = "ttc-lambda"
+    memory_limit_in_mb = 128
+    invoked_function_arn = "arn:aws:lambda:us-east-1:123456789012:function:ttc-lambda"
+    aws_request_id = "test-request-id"
```
nickclyde: `MockLambdaContext` is duplicated here, in `packages/augmentation-lambda/tests/test_augmentation_lambda_function.py`, and in `e2e/test_e2e.py`. Could we move it into `lambda-handler/test_constants` (or a shared `conftest.py` fixture) so the three copies don't drift? Not blocking, just a chance to DRY.

terraform/README.md
```diff
@@ -100,6 +100,65 @@ The augmentation Lambda uses only the Lambda security group (not the OpenSearch
...
+def handler(event, context):
+    ...
+```
+
+Shared/core packages called by Lambda code use Powertools child loggers rather than stdlib `logging`:
+
+```python
+from aws_lambda_powertools import Logger
+
+logger = Logger(service="ttc", child=True)
```
nickclyde: Could we add a one-liner here noting that `text-to-code` (and any future shared package using `child=True` loggers) requires the parent `Logger(service="ttc")` to be initialized for logs to flow at the expected level? It's a Powertools quirk that's easy to miss when running these helpers outside Lambda (e.g. local scripts in `data-curation` / `model-tuning`).

nickclyde (commented):
Awesome work, Rob! Just a few nits and ideas for drying it up, but thanks for tackling this.

packages/lambda-handler/pyproject.toml
robertandremitchell: yeah, I lean toward expanding to have on all the lambda functions

robertandremitchell: I can add them in this ticket

packages/text-to-code/pyproject.toml
robertandremitchell: That's a good point. I think for these, using logging would make sense. I can update the code.

packages/index-lambda/src/index_lambda/lambda_function.py
robertandremitchell: makes sense, will add

packages/text-to-code-lambda/tests/test_lambda_function.py
robertandremitchell: ahh good call out. yeah, I forgot we had the shared constants

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
robertandremitchell: removed!

packages/index-lambda/src/index_lambda/lambda_function.py
robertandremitchell: added!

packages/text-to-code-lambda/tests/test_lambda_function.py
robertandremitchell: moved this to the repo-wide conftest. I was trying it just in lambda-handler thinkign that would be "inherited" across all of them, but ended up needing to put the code there for it to work with the least duplication. curious your thoughts.

packages/lambda-handler/pyproject.toml
robertandremitchell: added!

packages/text-to-code/pyproject.toml
robertandremitchell: updated!

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
robertandremitchell: updated!

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
robertandremitchell: updated!

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py

terraform/README.md
```diff
@@ -100,6 +100,65 @@ The augmentation Lambda uses only the Lambda security group (not the OpenSearch
...
+```python
+logger.info(
+    "Retrieving eICR from S3",
+    bucket_name=bucket_name,
+    s3_key=object_key,
+    status="processing",
+)
+```
+
+Do not use `print()`, `structlog`, or stdlib `logging` in Lambda packages or shared packages used by Lambda code.
```
JNygaard-Skylight: There's a ruff rule that flags `print()` statements: [T201](https://docs.astral.sh/ruff/rules/print/). We could enable it for the relevant packages?

terraform/README.md
```diff
@@ -100,6 +100,65 @@ The augmentation Lambda uses only the Lambda security group (not the OpenSearch
 
 Environment variables injected at deploy time: `S3_BUCKET`, `AUGMENTED_EICR_PREFIX`, `AUGMENTATION_METADATA_PREFIX`, `REGION`.
 
+### Logging
```
JNygaard-Skylight, line 103: Does there need to be some logging documentation in the root README.md? It makes sense that the deployment docs would have a logging section, but some of this information is probably relevant to development in general. Idk, I do not feel strongly about making you write more documentation, so I'll leave it up to you, just wanted to bring it up.

robertandremitchell: I agree. I can add a section there, as well.

terraform/README.md
robertandremitchell: makes sense to me, can add!

robertandremitchell: updated!

terraform/README.md
robertandremitchell: updated!

nickclyde (approved):
Thanks for the changes Rob!!

PR #517, opened 2026-04-29
Add bulk testing script

### LINKED ISSUE #498: Create batch TTC pipeline testing script
## Description

Once a collection of test examples has been created for high-quality manual testing, we need a way to efficiently run the pipeline against each of those samples. Nick has done excellent work creating a local bash script capable of running a single example through the TTC pipeline, but we now want to generalize that so that it can run a test with multiple examples, collecting the results across all the separate runs. The purpose of this ticket is to write such a script.

## Acceptance Criteria

* A bash script is created in the `/scripts` directory that runs a battery of tests using a `.json` file as the source of its test examples
* The dependencies and usage of this script are documented in the `scripts/` `README` file
* The script can take structural inspiration from Nick's singleton script as far as the sections, structuring `gum`, accessing the AWS environment, etc. However, rather than print each trace of the tailing logs and the full output, we're interested only in the correct predictions. For each test case in the `.json` file, this bash script should output:

  - the nonstandard input
  - the expected standard output and LOINC code
  - the TTC pipeline predicted output string and predicted LOINC code
  - a string declaration of whether the answer was correct or incorrect

bamader:
## Description

This PR adds a bulk testing script to our repo that can be used in conjunction with a JSON file of test cases to submit an entire batch of test eICRs to the DIBBs pipeline and then pull down all of their results to check for accuracy.

## Related Issues

Closes #498 

## Additional Notes

I didn't use a whole lot of gum functionality but the script definitely works. It's just not as fashionable as Nick's lol.

### DIFF
--- a/scripts/bash_xml_parser.py
+++ b/scripts/bash_xml_parser.py
@@ -13,28 +13,28 @@
 #     that might require defusedxml (Billion Laughs, Quadratic Explosion)
 #  2. CPython's own documentation has removed recommendations to use defusedxml
 #  3. defusedxml is effectively a dead project, with no release in years and no
-#     official Python support for versions 3.9+, while the underlyling expat
+#     official Python support for versions 3.9+, while the underlying expat
 #     package continues to see hardening security fixes
 #  4. For purposes of bulk testing, we are fully in control of the XML data, since
-#     it originates from an eICR we send, which makes security overhead unneccessary.
+#     it originates from an eICR we send, which makes security overhead unnecessary.
 
 import os
+import sys
 import xml.etree.ElementTree as ET
 
 if __name__ == "__main__":
     xml = os.environ["CONTENT"]
 
-    root = ET.fromstring(xml)  # noqa
+    root = ET.fromstring(xml)
 
-    node = (
-        root.find("component")
-        .find("structuredBody")
-        .find("component")
-        .find("section")
-        .find("entry")
-        .find("observation")
-        .find("code")
-        .find("translation")
+    node = root.find(
+        "component/structuredBody/component/section/entry/observation/code/translation"
     )
+
+    if node is None:
+        print()
+        print()
+        sys.exit(0)
+
     print(node.get("code"))
     print(node.get("DisplayName"))
--- a/scripts/bash_xml_parser.py
+++ b/scripts/bash_xml_parser.py
@@ -1,0 +1,40 @@
+# bash_xml_parser.py
+#
+# Simple python helper script to parse an augmented eICR for its translated
+# LOINC code, which is where the TTC Pipeline inserts its predicted standardized
+# LOINC code.
+#
+# Note on use of XML Library: S314 recommends the use of the `defusedxml` library
+# over `xml` proper, on the grounds that it is more secure against some forms of
+# Denial of Service attacks and other security vulnerabilities. However, this rule
+# is largely out of date, and modern Python developments have sufficiently closed
+# the gap for several reasons:
+#  1. Python versions 3.11+ have built-in fixes to address the DoS vulnerabilities
+#     that might require defusedxml (Billion Laughs, Quadratic Explosion)
+#  2. CPython's own documentation has removed recommendations to use defusedxml
+#  3. defusedxml is effectively a dead project, with no release in years and no
+#     official Python support for versions 3.9+, while the underlyling expat
+#     package continues to see hardening security fixes
+#  4. For purposes of bulk testing, we are fully in control of the XML data, since
+#     it originates from an eICR we send, which makes security overhead unneccessary.
+
+import os
+import xml.etree.ElementTree as ET
+
+if __name__ == "__main__":
+    xml = os.environ["CONTENT"]
+
+    root = ET.fromstring(xml)  # noqa
+
+    node = (
+        root.find("component")
+        .find("structuredBody")
+        .find("component")
+        .find("section")
+        .find("entry")
+        .find("observation")
+        .find("code")
+        .find("translation")
+    )
+    print(node.get("code"))
+    print(node.get("DisplayName"))
--- a/scripts/batch_aws_test.sh
+++ b/scripts/batch_aws_test.sh
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 #
 # A batch-uploading bulk testing script for the DIBBs TTC Pipeline.
-#
+
 # Given a JSON file of nonstandard test cases, this script:
 #   1. Loads each test case from the file into a bash array.
-#   2. Templates a dummy test eicr with the nonstandard input name
+#   2. Templates a dummy test eICR with the nonstandard input name
 #      of each test case, and stamps each result with a fresh UUID.
 #   3. For each test eICR, one at a time, uploads the templated eICR + a 
 #      canned schematron errors file to S3, which fires the TTC lambda
@@ -15,26 +15,19 @@
 #      translated code name (where the TTC Pipeline leaves its predicted
 #      standardization), and compares the result to the expected value of
 #      the test case.
-#
-# Usage: ./scripts/aws_e2e.sh ./test_cases_file.json
+
+# Usage: ./scripts/batch_aws_test.sh ./test_cases_file.json
 #
 
 set -euo pipefail
 
 # ── Required tooling ──────────────────────────────────────────────────────────
 # gum:      TUI chrome (styled banners, spinners, log levels).
-# unbuffer: wraps `aws` in a pseudo-TTY so it line-buffers stdout. Without this
-#           the AWS CLI v2 (a PyInstaller bundle that ignores PYTHONUNBUFFERED)
-#           block-buffers output when piped and all log lines would arrive in
-#           one burst at the end.
 command -v gum >/dev/null || {
     echo "This script requires 'gum'. Install with: brew install gum" >&2
     exit 1
 }
-command -v unbuffer >/dev/null || {
-    echo "This script requires 'unbuffer' (from expect). Install with: brew install expect" >&2
-    exit 1
-}
+
 
 # ── Config ────────────────────────────────────────────────────────────────────
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
@@ -99,13 +92,13 @@ follow_until_report() {
 
     # Start the tail in the background; read its output from FD 3 so we can
     # use `read -t` on it (needs a persistent FD, not a redirect per-read).
-    unbuffer aws logs tail "$log_group" --follow --since 1m --color off >"$fifo" &
+    aws logs tail "$log_group" --region "$AWS_REGION" --follow --since "${start}s" --color off >"$fifo" &
     tail_pid=$!
     exec 3<"$fifo"
 
     while :; do
         if IFS= read -r -t 1 -u 3 line; then
-            :
+            [[ "$line" == *"REPORT RequestId:"* ]] && break
         else
             rc=$?
             (( rc > 128 )) || break   # rc > 128 = timeout; anything else = EOF
@@ -124,7 +117,7 @@ follow_until_report() {
 wait_for_s3_object() {
     local key="$1" label="$2" timeout="${3:-120}" interval=2 elapsed=0
 
-    while ! aws s3api head-object --bucket "$BUCKET" --key "$key" >/dev/null 2>&1; do
+    while ! aws s3api head-object --region "$AWS_REGION" --bucket "$BUCKET" --key "$key" >/dev/null 2>&1; do
         if (( elapsed >= timeout )); then
             gum log -l error "Timed out waiting for $label at s3://$BUCKET/$key"
             return 1
@@ -142,27 +135,16 @@ wait_for_s3_object() {
 # The test cases JSON file has some nested structuring to make it easy to
 # define new cases or modify existing ones, so the easiest way to programmatically
 # handle this is to use a simple helper script with a few lines of python that
-# loads the appropriate variable within the nested structure. We can call this
-# helper a few different times, giving a different variable name each time, as
-# this lets us avoid repeating any python code here and also avoids having to 
-# unpack JSON in bash (we get to skip directly to an array of strings).
-NONSTANDARD_INPUTS=()
-while IFS= read -r line; do
-    NONSTANDARD_INPUTS+=("$line")
-done < <(JSON_FP="$JSON_FP" python3 ./bash_json_loader.py "nonstandard_in")
-CORRECT_OUTPUTS=()
-while IFS= read -r line; do
-    CORRECT_OUTPUTS+=("$line")
-done < <(JSON_FP="$JSON_FP" python3 ./bash_json_loader.py "correct_standardized_code")
-LOINC_CODES=()
-while IFS= read -r line; do
-    LOINC_CODES+=("$line")
-done < <(JSON_FP="$JSON_FP" python3 ./bash_json_loader.py "numeric_loinc_code")
+# loads the appropriate variable within the nested structure.
+NONSTANDARD_INPUTS=(); CORRECT_OUTPUTS=(); LOINC_CODES=()
+while IFS=$'\t' read -r nin cout loinc; do
+    NONSTANDARD_INPUTS+=("$nin"); CORRECT_OUTPUTS+=("$cout"); LOINC_CODES+=("$loinc")
+done < <(JSON_FP="$JSON_FP" python3 "$SCRIPT_DIR/bash_json_loader.py")
 
 for i in "${!NONSTANDARD_INPUTS[@]}"; do
 
     section "Test Case $i:"
-    FILENAME="$(date +%m-%d-%Y_%H:%M:%S).xml"
+    FILENAME="$(date +%m-%d-%Y_%H:%M:%S)_$i.xml"
     TEMPLATED_EICR="$TMPDIR/$FILENAME"
 
     # ── Template the eICR ─────────────────────────────────────────────────────────
@@ -171,7 +153,7 @@ for i in "${!NONSTANDARD_INPUTS[@]}"; do
     # <id>/<setId> so this eICR looks like a new document to downstream systems.
     echo "  Templating eICR with nonstandard input '${NONSTANDARD_INPUTS[$i]}'"
     INPUT="${NONSTANDARD_INPUTS[$i]}" SOURCE_EICR="$SOURCE_EICR" OUT_PATH="$TEMPLATED_EICR" \
-        python3 ./bash_eicr_templater.py
+        python3 "$SCRIPT_DIR/bash_eicr_templater.py"
     
     
     # # ── Fire the pipeline ─────────────────────────────────────────────────────────
@@ -181,9 +163,9 @@ for i in "${!NONSTANDARD_INPUTS[@]}"; do
     # # the same filename.
     echo "  Uploading to S3"
     gum spin --spinner=dot --title "Uploading schematron errors…" -- \
-        aws s3 cp "$SOURCE_SCHEMATRON" "s3://$BUCKET/ValidationResponseV2/$FILENAME"
+        aws s3 cp --region "$AWS_REGION" "$SOURCE_SCHEMATRON" "s3://$BUCKET/ValidationResponseV2/$FILENAME"
     gum spin --spinner=dot --title "Uploading templated eICR…" -- \
-        aws s3 cp "$TEMPLATED_EICR" "s3://$BUCKET/TextToCodeSubmissionV2/$FILENAME"
+        aws s3 cp --region "$AWS_REGION" "$TEMPLATED_EICR" "s3://$BUCKET/TextToCodeSubmissionV2/$FILENAME"
 
     # Block until each lambda finishes. TTC writes TTCAugmentationMetadataV2/<name>.json,
     # which triggers the augmentation lambda; that's why we watch them in sequence.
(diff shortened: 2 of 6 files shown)


robertandremitchell (commented):
is there a dummy file or local file we can create/use for testing purposes?

bamader:
@robertandremitchell Yes! The `test_cases.json` file in the same `scripts/` directory as this. As with Nick's script, make sure you're logged in / authenticated to AWS before running.

robertandremitchell:
> @robertandremitchell Yes! The `test_cases.json` file in the same `scripts/` directory as this. As with Nick's script, make sure you're logged in / authenticated to AWS before running.

ack, the golden retrievers have stolen my brain cells 🤦🏽

scripts/batch_aws_test.sh
```diff
@@ -0,0 +1,210 @@
...
+#      canned schematron errors file to S3, which fires the TTC lambda
+#      via SQS event.
+#   4. Tails both lambdas' CloudWatch logs in real time until each emits its
+#      `REPORT RequestId:` line (AWS's end-of-invocation marker).
+#   5. Fetches the resulting augmented eICR XML from S3, parses its
+#      translated code name (where the TTC Pipeline leaves its predicted
+#      standardization), and compares the result to the expected value of
+#      the test case.
+#
+# Usage: ./scripts/aws_e2e.sh ./test_cases_file.json
```
nickclyde: Tiny doc nit: this still says `./scripts/aws_e2e.sh` from the original. I think it should be `./scripts/batch_aws_test.sh ./test_cases_file.json`.

scripts/batch_aws_test.sh
```diff
@@ -0,0 +1,210 @@
...
+        fi
+    done
+
+    exec 3<&-
+    kill "$tail_pid" 2>/dev/null || true
+    wait "$tail_pid" 2>/dev/null || true
+    rm -f "$fifo"
+
+    echo "  $label complete in $(( $(date +%s) - start ))s"
+}
```
nickclyde, line 114: I think this loop will hang forever once a lambda fires (actually just confirmed by running locally). In the original `aws_e2e.sh` the success branch had `[[ "$line" == *"REPORT RequestId:"* ]] && break` - that check seems to have been dropped here, and the success branch is now just `:`. Without it, the only exit condition is EOF on the FIFO, but `aws logs tail --follow` doesn't EOF on its own.

A second issue once that's fixed: `--since 1m` on line 102 will pick up the previous test case's `REPORT RequestId:` line on iteration 2+, so each subsequent case is likely to break out instantly before the lambda has even started. A couple of options:
  - Capture `start=$(date +%s)` (already there) and use `--since "${start}s"` or skip lines with timestamps earlier than `start`.
  - Track each invocation's RequestId from a known marker (e.g. by tagging the upload with a unique key and grepping for that) instead of the generic `REPORT` line.

scripts/batch_aws_test.sh
```diff
@@ -0,0 +1,210 @@
...
+#           block-buffers output when piped and all log lines would arrive in
+#           one burst at the end.
+command -v gum >/dev/null || {
+    echo "This script requires 'gum'. Install with: brew install gum" >&2
+    exit 1
+}
+command -v unbuffer >/dev/null || {
+    echo "This script requires 'unbuffer' (from expect). Install with: brew install expect" >&2
+    exit 1
+}
```
nickclyde: I don't think `unbuffer` is needed anymore - it was there in `aws_e2e.sh` so `pretty_log_line` could render lines as they arrived. This script seems to no longer prints log lines (the read body is `:`), so the buffering behavior of `aws logs tail` doesn't matter. Probably safe to drop both the requirement check and the `unbuffer` wrapper on line 102.

scripts/batch_aws_test.sh
```diff
@@ -0,0 +1,210 @@
...
+
+# ── Config ────────────────────────────────────────────────────────────────────
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+BUCKET="dibbs-text-to-code"
+SOURCE_EICR="$SCRIPT_DIR/test_eicr.xml"
+SOURCE_SCHEMATRON="$SCRIPT_DIR/test_schematron_errors.xml"
+TTC_LOG="/aws/lambda/ttc-lambda"                  # CloudWatch log group for the TTC lambda
+AUG_LOG="/aws/lambda/ttc-augmentation-lambda"     # …and the augmentation lambda
+AWS_REGION="us-east-2"
```
nickclyde, line 39: `AWS_REGION` is set but I think the AWS CLI calls fall back to `~/.aws/config` or the ambient `AWS_REGION` env var. If we want to pin region explicitly, we'd need to prefix the AWS calls with `AWS_REGION="$AWS_REGION"` or run `aws --region "$AWS_REGION" …`. I think this would need to change in aws_e2e.sh as well.

scripts/batch_aws_test.sh
```diff
@@ -0,0 +1,210 @@
...
+    NONSTANDARD_INPUTS+=("$line")
+done < <(JSON_FP="$JSON_FP" python3 ./bash_json_loader.py "nonstandard_in")
+CORRECT_OUTPUTS=()
+while IFS= read -r line; do
+    CORRECT_OUTPUTS+=("$line")
+done < <(JSON_FP="$JSON_FP" python3 ./bash_json_loader.py "correct_standardized_code")
+LOINC_CODES=()
+while IFS= read -r line; do
+    LOINC_CODES+=("$line")
+done < <(JSON_FP="$JSON_FP" python3 ./bash_json_loader.py "numeric_loinc_code")
```
nickclyde: Three calls into `bash_json_loader.py` re-parse the file every time and silently break if the three keys don't all have the same number of values. I think it'd be safer to load once and emit all three columns at once, e.g. tab-separated, then read into the three arrays in a single pass:

```bash
NONSTANDARD_INPUTS=(); CORRECT_OUTPUTS=(); LOINC_CODES=()
while IFS=$'\t' read -r nin cout loinc; do
  NONSTANDARD_INPUTS+=("$nin"); CORRECT_OUTPUTS+=("$cout"); LOINC_CODES+=("$loinc")
done < <(JSON_FP="$JSON_FP" python3 "$SCRIPT_DIR/bash_json_loader.py")
```

That also lets the loader validate per-row that all three keys exist before emitting anything.

scripts/batch_aws_test.sh
```diff
@@ -0,0 +1,210 @@
...
+# define new cases or modify existing ones, so the easiest way to programmatically
+# handle this is to use a simple helper script with a few lines of python that
+# loads the appropriate variable within the nested structure. We can call this
+# helper a few different times, giving a different variable name each time, as
+# this lets us avoid repeating any python code here and also avoids having to 
+# unpack JSON in bash (we get to skip directly to an array of strings).
+NONSTANDARD_INPUTS=()
+while IFS= read -r line; do
+    NONSTANDARD_INPUTS+=("$line")
+done < <(JSON_FP="$JSON_FP" python3 ./bash_json_loader.py "nonstandard_in")
```
nickclyde: I believe the `./bash_json_loader.py` invocations on lines 152/156/160 (and `./bash_eicr_templater.py` on 174, `./bash_xml_parser.py` on 198) only resolve when the script is run from the `scripts/` directory. `SCRIPT_DIR` is already computed up top - could you switch these to `python3 "$SCRIPT_DIR/bash_json_loader.py"` etc. so the script works from any cwd? The header comment example (`./scripts/aws_e2e.sh ./test_cases_file.json`) suggests it should be runnable from the repo root.

scripts/batch_aws_test.sh
```diff
@@ -0,0 +1,210 @@
...
+done < <(JSON_FP="$JSON_FP" python3 ./bash_json_loader.py "correct_standardized_code")
+LOINC_CODES=()
+while IFS= read -r line; do
+    LOINC_CODES+=("$line")
+done < <(JSON_FP="$JSON_FP" python3 ./bash_json_loader.py "numeric_loinc_code")
+
+for i in "${!NONSTANDARD_INPUTS[@]}"; do
+
+    section "Test Case $i:"
+    FILENAME="$(date +%m-%d-%Y_%H:%M:%S).xml"
```
nickclyde: Since `FILENAME` only goes to second precision, two iterations that happen to land in the same second would collide on the same S3 key (and on disk in `$TMPDIR`), which would silently overwrite the in-flight case. Once `follow_until_report` waits properly this is unlikely to happen in practice (each iteration takes seconds), but a counter or `%N`-equivalent suffix (`$(date +%s%N)` or appending `_$i`) would make this robust against future parallelization too.

scripts/batch_aws_test.sh
```diff
@@ -0,0 +1,210 @@
...
+    echo "  Fetching results..."
+    wait_for_s3_object "AugmentationEICRV2/$FILENAME" "augmented eICR"
+    content="$(aws s3 cp "s3://$BUCKET/AugmentationEICRV2/$FILENAME" -)"
+    { read PREDICTED_LOINC; read PREDICTED_CODE_STRING; } < <(CONTENT="$content" python3 ./bash_xml_parser.py)
+
+    if [[ "$PREDICTED_LOINC" = "${LOINC_CODES[$i]}" ]]; then
+        echo "  Predicted LOINC code $PREDICTED_LOINC matches expected code ${LOINC_CODES[$i]}"
+        echo "  Test Case Passed!"
+    else
+        echo "  Predicted code $PREDICTED_LOINC does not match expected code "${LOINC_CODES[$i]}""
```
nickclyde: Quoting nit: the `"${LOINC_CODES[$i]}""` here closes the outer double quote partway through and re-opens it, so this only works because LOINC codes don't contain spaces. I think you want either:

```bash
echo "  Predicted code $PREDICTED_LOINC does not match expected code ${LOINC_CODES[$i]}"
```

or consistently escape: `\"${LOINC_CODES[$i]}\"` if you're trying to render literal quotes around the value.

scripts/bash_xml_parser.py
```diff
@@ -0,0 +1,40 @@
...
+        .find("structuredBody")
+        .find("component")
+        .find("section")
+        .find("entry")
+        .find("observation")
+        .find("code")
+        .find("translation")
+    )
+    print(node.get("code"))
+    print(node.get("DisplayName"))
```
nickclyde: If TTC fails to predict a code (so the augmenter never inserts a `<translation>`), any of these `find()` calls returns `None` and you get an `AttributeError` rather than a clean "no prediction" outcome. Combined with `set -euo pipefail` in the bash driver, that aborts the whole batch on the first un-predicted case. Could we guard with something like:

   ```python
   node = root.find("component/structuredBody/component/section/entry/observation/code/translation")
   if node is None:
       print("")
       print("")
       sys.exit(0)
   ```

   so the bash side can detect the empty prediction and report it as a failed case?

nickclyde (commented):
Thanks for getting this together Brandon! Having a way to run a batch of cases against the deployed pipeline will be really useful for tracking accuracy regressions over time. The overall flow (template, upload, wait, parse, compare) is clear and the helper python scripts are nicely scoped.

A few notes below, but overall looking good!

A broader thought: since `aws_e2e.sh` and `batch_aws_test.sh` now share ~80% of their plumbing (the templating, the S3 upload sequencing, the log tail, the tooling preamble), it might be worth either factoring the shared bits into a library/utils file that both scripts source, or having `aws_e2e.sh` just delegate to `batch_aws_test.sh` with a single-element JSON. Not blocking — just flagging because every drift between the two scripts is going to mean fixing things twice. Happy to take a swing at that as a follow-up if helpful.

A `scripts/README.md` update with how to author a new test case file (the JSON schema, where to put it, expected output) would also help the next person.

scripts/bash_xml_parser.py
```diff
@@ -0,0 +1,40 @@
+# bash_xml_parser.py
+#
+# Simple python helper script to parse an augmented eICR for its translated
+# LOINC code, which is where the TTC Pipeline inserts its predicted standardized
+# LOINC code.
+#
+# Note on use of XML Library: S314 recommends the use of the `defusedxml` library
```
JNygaard-Skylight, line 7: Should this rule be added to the excluded rule list?

scripts/batch_aws_test.sh
```diff
@@ -0,0 +1,210 @@
+#!/usr/bin/env bash
+#
+# A batch-uploading bulk testing script for the DIBBs TTC Pipeline.
+#
+# Given a JSON file of nonstandard test cases, this script:
+#   1. Loads each test case from the file into a bash array.
+#   2. Templates a dummy test eicr with the nonstandard input name
```
JNygaard-Skylight: Minute nit:
```suggestion
#   2. Templates a dummy test eICR with the nonstandard input name
```

scripts/bash_xml_parser.py
```diff
@@ -0,0 +1,40 @@
...
+# over `xml` proper, on the grounds that it is more secure against some forms of
+# Denial of Service attacks and other security vulnerabilities. However, this rule
+# is largely out of date, and modern Python developments have sufficiently closed
+# the gap for several reasons:
+#  1. Python versions 3.11+ have built-in fixes to address the DoS vulnerabilities
+#     that might require defusedxml (Billion Laughs, Quadratic Explosion)
+#  2. CPython's own documentation has removed recommendations to use defusedxml
+#  3. defusedxml is effectively a dead project, with no release in years and no
+#     official Python support for versions 3.9+, while the underlyling expat
```
JNygaard-Skylight:
```suggestion
#     official Python support for versions 3.9+, while the underlying expat
```

scripts/bash_xml_parser.py
```diff
@@ -0,0 +1,40 @@
...
+#     package continues to see hardening security fixes
+#  4. For purposes of bulk testing, we are fully in control of the XML data, since
+#     it originates from an eICR we send, which makes security overhead unneccessary.
```
JNygaard-Skylight:
```suggestion
#     it originates from an eICR we send, which makes security overhead unnecessary.
```

scripts/batch_aws_test.sh
bamader: Good catch, I must've been overzealous removing some of the gum pretty log printing. I've re-added the break conditional and have switched to `--since "${start}s" and the `follow_until_report` seems to be working well. Once I push the fix up let me know if you still have a hanging situation during upload.

scripts/batch_aws_test.sh
bamader: Yeah since we were testing 10+ cases in this script I thought printing every log line would be too cluttered (that's why I clarified the other day that your e2e smoke test script is intended for full end to end functionality, this one is more about assessing results and getting a batch of matches). I'll go ahead and remove unbuffer so we don't have the clutter.

scripts/batch_aws_test.sh
bamader: Ah another good catch, I think I configured my local config to automatically use the region the script is hitting. I'll add the prefixing so that the script doesn't overwrite anyone's configs but everything still lands appropriately!

scripts/batch_aws_test.sh
bamader: This is 100% a place where my bash-fu weakness really came through 😅 haha. Thanks for the helpful code for consolidating into a single pass, this will absolutely be much cleaner and more effective!

robertandremitchell (commented):
I think there's some filepathing that maybe needs to be more cleanly resolved. was puzzling for a second where I was getting this error unless I ran it from directly inside `scripts`


```sh
(dibbs-text-to-code-root) rob@Mac dibbs-text-to-code % bash scripts/batch_aws_test.sh scripts/test_cases.json
                                                 
╭───────────────────────────────────────────────╮
│  DIBBs Text to Code Batch Pipeline Test       │
│                                               │
│  File of Test Cases: scripts/test_cases.json  │
│  Bucket:    s3://dibbs-text-to-code           │
╰───────────────────────────────────────────────╯
                                                 
/opt/homebrew/Cellar/python@3.11/3.11.13/Frameworks/Python.framework/Versions/3.11/Resources/Python.app/Contents/MacOS/Python: can't open file '/Users/rob/dibbs-text-to-code/./bash_json_loader.py': [Errno 2] No such file or directory
/opt/homebrew/Cellar/python@3.11/3.11.13/Frameworks/Python.framework/Versions/3.11/Resources/Python.app/Contents/MacOS/Python: can't open file '/Users/rob/dibbs-text-to-code/./bash_json_loader.py': [Errno 2] No such file or directory
/opt/homebrew/Cellar/python@3.11/3.11.13/Frameworks/Python.framework/Versions/3.11/Resources/Python.app/Contents/MacOS/Python: can't open file '/Users/rob/dibbs-text-to-code/./bash_json_loader.py': [Errno 2] No such file or directory
```

scripts/batch_aws_test.sh
bamader: Yep, fully a fat-finger there, definitely didn't want the middle set of quotes at all

scripts/bash_xml_parser.py
bamader: I would say yes, but I didn't want to unilaterally make that call. If the rationale I outlined here makes sense, though, I'm happy to add S314 to our global project exclusions!

bamader:
@robertandremitchell You're 100% right on the file pathing, Nick observed that I wasn't prefixing the `python3` calls with the `$SCRIPT_DIR` we established, so that fix will be included in the next commit and things should now work from any directory. 

Sounds like that aside, you were able to run the script from the scripts directory without hanging waiting for logs?

bamader:
Readme is updated and all comments and suggestions should be addressed!

robertandremitchell:
> @robertandremitchell You're 100% right on the file pathing, Nick observed that I wasn't prefixing the `python3` calls with the `$SCRIPT_DIR` we established, so that fix will be included in the next commit and things should now work from any directory.
> 
> Sounds like that aside, you were able to run the script from the scripts directory without hanging waiting for logs?

yep, no other issues on my end!

robertandremitchell (approved):
took about a minute total for me, just confirming we expect some fails?

```
(dibbs-text-to-code-root) rob@Mac dibbs-text-to-code % bash scripts/batch_aws_test.sh scripts/test_cases.json
                                                 
╭───────────────────────────────────────────────╮
│  DIBBs Text to Code Batch Pipeline Test       │
│                                               │
│  File of Test Cases: scripts/test_cases.json  │
│  Bucket:    s3://dibbs-text-to-code           │
╰───────────────────────────────────────────────╯
                                                 
              
▸ Test Case 0:
  Templating eICR with nonstandard input 'blood urea nitrogen (BUN)'
  Uploading to S3
  TTC complete in 1s
  Augmentation complete in 1s
  Fetching results...
  Predicted LOINC code 6299-2 matches expected code 6299-2
  Test Case Passed!
              
▸ Test Case 1:
  Templating eICR with nonstandard input 'K+, Whole Blood'
  Uploading to S3
  TTC complete in 1s
  Augmentation complete in 1s
  Fetching results...
  Predicted code 75940-7 does not match expected code 6298-4
    Predicted code string: Potassium [Mass/volume] in Blood
    Expected code string:  Potassium [Moles/volume] in Blood
  Test Case Failed :/
              
▸ Test Case 2:
  Templating eICR with nonstandard input 'POC Oxymorphone, Qualitative, Urine'
  Uploading to S3
  TTC complete in 1s
  Augmentation complete in 1s
  Fetching results...
  Predicted LOINC code 11247-4 matches expected code 11247-4
  Test Case Passed!
              
▸ Test Case 3:
  Templating eICR with nonstandard input 'Calculated VLDL'
  Uploading to S3
  TTC complete in 1s
  Augmentation complete in 1s
  Fetching results...
  Predicted code 41999-4 does not match expected code 13458-5
    Predicted code string: VLDL Fld Ql
    Expected code string:  Cholesterol in VLDL Calc [Mass/vol]
  Test Case Failed :/
              
▸ Test Case 4:
  Templating eICR with nonstandard input 'pCO2 Venous'
  Uploading to S3
  TTC complete in 1s
  Augmentation complete in 1s
  Fetching results...
  Predicted LOINC code 2021-4 matches expected code 2021-4
  Test Case Passed!
              
▸ Test Case 5:
  Templating eICR with nonstandard input 'hep A Ab, IgM'
  Uploading to S3
  TTC complete in 2s
  Augmentation complete in 1s
  Fetching results...
  Predicted code 22314-9 does not match expected code 22315-6
    Predicted code string: Hepatitis A virus IgM Ab [Presence] in Serum
    Expected code string:  Hepatitis A virus IgM Ab [Units/volume] in Serum
  Test Case Failed :/
              
▸ Test Case 6:
  Templating eICR with nonstandard input 'Automated erythrocyte mean corpuscular hemoglobin concentration (MCHC) measurement (m'
  Uploading to S3
  TTC complete in 1s
  Augmentation complete in 1s
  Fetching results...
  Predicted LOINC code 786-4 matches expected code 786-4
  Test Case Passed!
              
▸ Test Case 7:
  Templating eICR with nonstandard input 'Screening urine fentanyl detection'
  Uploading to S3
  TTC complete in 1s
  Augmentation complete in 1s
  Fetching results...
  Predicted LOINC code 59673-4 matches expected code 59673-4
  Test Case Passed!
              
▸ Test Case 8:
  Templating eICR with nonstandard input 'GLUCOSE 1H POST 50 G GLUCOSE PO'
  Uploading to S3
  TTC complete in 1s
  Augmentation complete in 1s
  Fetching results...
  Predicted LOINC code 1504-0 matches expected code 1504-0
  Test Case Passed!
              
▸ Test Case 9:
  Templating eICR with nonstandard input 'CBC Pnl (Auto Diff), Bld'
  Uploading to S3
  TTC complete in 1s
  Augmentation complete in 1s
  Fetching results...
  Predicted code 32207-3 does not match expected code 57021-8
    Predicted code string: PDW Bld Auto
    Expected code string:  CBC W Auto Differential panel (Bld)
  Test Case Failed :/
```

bamader:
Yep, I believe we expect 4 failures out of our 10. These test cases are actual production-grade samples that we're measuring performance against, so failure here just indicates a non-match. That all looks as expected to me!

scripts/bash_xml_parser.py
JNygaard-Skylight: Makes sense to me!

nickclyde (approved):
Awesome work Brandon!!

PR #521, opened 2026-05-01
refactor: Remove unused FastAPI service and React frontend

nickclyde:
## Summary

- The FastAPI service (`packages/api`) and React frontend (`frontend/`) were a proof-of-concept demo and are no longer in use. The active surface of this repo is the AWS Lambda pipeline plus its supporting Terraform infrastructure.
- Deletes `frontend/`, `packages/api/`, `Dockerfile.app`, root `package.json`/`package-lock.json`, prettier config, the `dev` just submodule, and the `demo-site`/`api-service` docker-compose services.
- Trims supporting config: removes the prettier pre-commit hook, the npm dependabot ecosystem, the `packages/api` workspace member and ty/ruff entries, the `FAST` ruff rule selection, and the now-stale `frontend/` lines from `.dockerignore`.

## Test plan

- [x] `uv lock` regenerates cleanly (drops fastapi, uvicorn, starlette, sentry-sdk, watchfiles, websockets, etc.)
- [x] `uv sync --all-packages` succeeds
- [x] `uv run ruff check` — All checks passed
- [x] `uv run ty check` — All checks passed
- [x] `pre-commit run --all-files` — all hooks pass without the prettier hook
- [x] `pytest --collect-only` — 308 tests collected
- [x] `docker compose config` — trimmed file parses; `lambda` and `opensearch-node` services remain
- [ ] CI passes


.pre-commit-config.yaml
```diff
@@ -19,11 +19,6 @@ repos:
         args: [--fix]
       - id: ruff-format # Run the formatter
 
-  - repo: https://github.com/JoC0de/pre-commit-prettier
```
JNygaard-Skylight, line 22: Prettier also formats Markdown, YAML, and JSON, so we should keep this hook.

nickclyde: Good catch — restored the prettier hook along with `.prettierrc` and `.prettierignore`. Dropped the `prettier-plugin-tailwindcss` plugin from `.prettierrc` since the frontend (and its tailwind usage) is gone in this PR.

JNygaard-Skylight (approved):
🪓

.pre-commit-config.yaml
JNygaard-Skylight: Perhaps there should be a comment in `.pre-commit-config.yaml` listing the files types we expect prettier to format? We could make it explicit and specify the file types in the hook with `type: ['json', 'markdown', 'yaml']`, but the default behaviour is to run for all types, and I would not want to have to remember to add types here.  Maybe there should be some details on what each pre-commit hook does in the readme?

This is only tangentially related to the original purpose of the PR so I don't think you have to do anything for this PR, but I wanted to call it out.

PR #523, opened 2026-05-04
bug: Get the top reranked hit

### LINKED ISSUE #510: TTC Lambda Function is returning wrong standardization to Augmentation
The Text to Code lambda function uses both a Retriever model and a Reranker model to select the best candidate out of the results returned by OpenSearch. The correct behavior of the lambda is that the OpenSearch results are re-scored and re-sorted by our Reranker before the highest scoring re-ranker selection is appended to the TTC Output. However, this is not happening in our codebase. Currently, we are choosing the highest scoring LOINC code from the _Retriever_ step, which operates only the query results returned by OpenSearch. The scope of this ticket is to correct this bug.

## Acceptance Criteria:

- In `text-to-code-lambda/src/text-to-code-lambda/lambda_function.py`, the code in the function `process_schematron_errors` is updated to use the reranker results rather than the retriever results. The function begins approximately on line 235, and the block of code in which the particular code is appended to TTC output is on approximately line 310. The values passed in to the `_build_nonstandard_code_instance` function call on approximately line 312 should use the code strings and values from the `ranked_results` list, not the `results_list`.
- Any applicable tests are updated to reflect this change

JNygaard-Skylight:
## Description

The TTC lambda was returning the highest ranked hit from the initial Opensearch query, but it should be using the highest ranked, after reranking.

## Related Issues

Closes #510 

## Additional Notes

Because I was there, I also added type hinting to the reranker unit tests.


packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -330,15 +330,20 @@ def _process_schematron_errors(
         retrieved_loinc_names = [hit.source.description for hit in results_list]
         ranked_results = rerank(selected_candidate.value, retrieved_loinc_names)
 
-        if results_list:
+        top_result = next(
+            (x for x in results_list if x.source.description == ranked_results[0]["code_string"]),
+            None,
+        )
```
nickclyde, line 336: I think this introduces a regression for the empty-hits case. If `results_list` is empty (OpenSearch returned no hits), then `retrieved_loinc_names` is `[]`, `rerank(...)` returns `[]`, and `ranked_results[0]["code_string"]` on line 334 will raise `IndexError` before the `if top_result:` guard on line 338 can run. The previous `if results_list:` check guarded against this, and the `else` branch / `if not results_list` block below was the intended unmatched-error path.

Could we re-introduce the empty-results guard before computing `top_result`? Something like:

```python
if not results_list:
    # fall through to the unmatched-error branch below
    ...
else:
    top_result = next(
        (x for x in results_list if x.source.description == ranked_results[0]["code_string"]),
        None,
    )
    if top_result:
        ...
```

nickclyde, line 336: Random thought: matching the top reranked entry back to its `OpenSearchHit` by string-comparing `description` works, but it is a bit fragile - two hits with the same description are rare but possible, e.g. if OpenSearch ever returns near-duplicates. It would silently pick whichever appears first in `results_list`, even if the reranker scored them differently.

Since `_RERANKER.rank(...)` already returns `corpus_id` indices internally, would it be worth threading a `corpus_id` (or `index`) through `ScoredResult` so we can do `top_result = results_list[ranked_results[0]["corpus_id"]]` directly? Maybe this is more of a follow-up ticket thing, just flagging it.

nickclyde (commented):
Thanks for tackling this bug, Josh, and for the bonus type hints in the reranker tests! Just a couple notes:

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
JNygaard-Skylight: This makes sense to me, however this scenario already has a test: `test_handler_adds_unmatched_error_when_selected_candidate_has_no_opensearch_hits`. It looks like it is defaulting to the given default value `None`. To be honest, I did not think the default value would kick in for any error, but I think this behaviour works for us?

JNygaard-Skylight: This makes sense to me. I think it is relevant and shouldn't add too much more to this PR.

JNygaard-Skylight: I reverted the changes I made for this. It ended up requiring a bunch of smaller changes with how the metadata and Lambda function output objects were constructed. I'll create a new ticket for this.

BradySkylight: Question about this.  We technically shouldn't get 'two hits with the same description' where the loinc codes are different, unless I'm not understanding something.

nickclyde: Ahhh. I see now. The `if` condition only evaluates if the iterator has something to iterate on, which it doesn't if results_list is `[]`. Okay, I think we're good then, glad we have a test for it.

nickclyde (approved):
Thanks Josh!

PR #532, opened 2026-05-06
Create functionality to extract LOINC Delta updates to generate additional embeddings

### LINKED ISSUE #452: Create Process to get delta update for valueset for LOINC

BradySkylight:
## Description

This PR includes changes to make it possible to perform updates against, at least LOINC for now, various medical terminology value-sets utilized by TTC.  Below is a list of the changes you will see:

- Another refactor that moves the common functionality for All value-sets (general.py) as well as specific functionality for each of the different value-sets/ontologies into their own `util` python script.  
- The old valuset_sync script has been renamed to 'extraction' and leverages the utils functionality to perform a FULL extraction and saves the various files
- The readme has been updated to indicate naming changes and location changes for various files/functionality
- The embedding notebook was also changed to reflect the new order of all LOINC data pulled from LOINC extract so it will work correctly if ever used again.
   - **NOTE:** There are some formatting changes in this file for some reason and I'm not sure why - ignore these unless you notice an issue
- Added a maintenance.py to handle maintaining the valueset extracts and embeddings for all valuesets - though only LOINC has been implemented so far for `loinc lab names`, as this is the only valueset loaded in our model at this point.

## Related Issues

Closes #452 

## Additional Notes

I haven't added in tests, but have tested along the way and have some print statements included for now just to see the end result.  The output will be passed to the next step - create the embeddings - which will then be stored as files somewhere and then uploaded into OpenSearch.

## Checklist for Reviewers


packages/data-curation/src/data_curation/terminologies/utils/general.py
```diff
@@ -0,0 +1,60 @@
...
+
+# File & Directories
+import csv
+from datetime import datetime
+import os
+import re
+
+
+# Value Set Directories
+SNOINC_DIRECTORY = "./data/snoinc_extracts"
```
robertandremitchell: more general feedback on different points of these directories / folders, we should use an OS-agnostic initialization akin to what the `validation` works does here: https://github.com/CDCgov/dibbs-text-to-code/blob/main/packages/validation/src/validation/main.py#L7-L21

packages/data-curation/src/data_curation/terminologies/general.py
```diff
@@ -0,0 +1,60 @@
...
+
+# regex patterns
+MULTIPLE_SPACE = re.compile(r"\s+")
+
+# Keys - the UMLS key is used for more than 
+#   one terminology set so it's here in General
+UMLS_API_KEY = os.environ.get("UMLS_API_KEY")
+
+
+def clean_text_string(value: str) -> str:
```
robertandremitchell, line 31: also more general, but we should have doc strings on these imo

packages/data-curation/src/data_curation/terminologies/hl7.py
```diff
@@ -0,0 +1,112 @@
...
+HL7_ENCOUNTER_CODE_URL = "https://terminology.hl7.org/6.5.0/CodeSystem-v3-ActCode.json"
+
+
+def get_hl7_encounter_act_codes():  # noqa: D103
+    hl7_response = requests.get(HL7_ENCOUNTER_CODE_URL)
+    encounter_act_code = "_ActEncounterCode"
+    hl7_rows = []
+
+    if hl7_response.status_code != 200:
+        print(
```
robertandremitchell, line 36: if we do envision this being pulled into the index-lambda or another service, we should use the `logging` library over printing

robertandremitchell (commented):
looks good! my concerns are mostly about documentation and aligning code with the other `packages` modules.

packages/data-curation/src/data_curation/terminologies/utils/general.py
BradySkylight: I can try to implement something like that.  I do like/prefer the use of os vs. other packages though.  I'll address this in the next PR, as it's in flight now for the embeddings.

packages/data-curation/src/data_curation/terminologies/general.py
BradySkylight: Can add those to all the functions for sure.

packages/data-curation/src/data_curation/terminologies/hl7.py
BradySkylight: Agreed.  This can be tackled in the next PR(s) as that is when we will start to wrap it in a lambda.  For now, it's nice to have a way to see progress when running locally.  But I'll add TODOs for each 'print' to convert to a logging statement instead.

robertandremitchell: sounds good!

packages/data-curation/src/data_curation/terminologies/general.py
JNygaard-Skylight: When I expanded the ruff rules I excluded several directories so the PR did not get too big, including `data-curation`. My thinking was that we would remove the exclusions as we worked on them.  I don't think it should be with this PR as it will add a lot of extra noise to the PR, but there should be a follow-up ticket to remove `data-curation` from the ruff exclude list and resolve any issues that come up.

BradySkylight: Agreed.  I was thinking we may move some of this functionality out of data_curation, possibly, and into a closer location to the lambda it will live in.  So, the goal may be accomplished by this as well.  Either way in another PR for sure.

packages/data-curation/src/data_curation/terminologies/general.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python"""
+
+"""
+data_curation.terminologies.utils.general
```
bamader, line 4: _Love_ this new sub-module directory. That's really going to help keep things organized and might even let us consolidate and refactor down the road!

packages/data-curation/src/data_curation/terminologies/utils/general.py
```diff
@@ -0,0 +1,100 @@
...
+import re
+
+
+# Value Set Directories
+BASE_FOLDER = Path(__file__).parents[4] / "data" / "snoinc_extracts"
+ENHANCEMENTS_DIRECTORY = BASE_FOLDER / "enhancements"
+TMP_DIRECTORY = "tmp"
+
+# regex patterns
+MULTIPLE_SPACE = re.compile(r"\s+")
```
bamader: Non-blocking, but I think our regex patterns util has a multiple-space one already pre-compiled

packages/data-curation/src/data_curation/terminologies/general.py
```diff
@@ -0,0 +1,100 @@
...
+        :param filename_prefix: The part of the filename that defines the
+            terminology value set type (ie. loinc_lab_names) that you want
+            to find the most recent file of.
+
+        :returns: The file name and file path of the most recent value set
+            extract file.
+    """
+    files = [f for f in os.listdir(BASE_FOLDER) if f.startswith(filename_prefix)]
+    if files:
+        latest_file = max(files)
```
bamader, line 91: Oh nice, very clever!

packages/data-curation/src/data_curation/terminologies/utils/hl7.py
```diff
@@ -0,0 +1,131 @@
...
+
+# Set Terminology URLS
+HL7_LAB_INTERP_URL = (
+    "https://terminology.hl7.org/2.1.0/CodeSystem-v3-ObservationInterpretation.json"
+)
+HL7_ENCOUNTER_CODE_URL = "https://terminology.hl7.org/6.5.0/CodeSystem-v3-ActCode.json"
+
+
+def get_hl7_encounter_act_codes() -> list[dict]:
+    """Function to get the all HL7 Codes and Terms for
```
bamader: Nit: Should be "all the," right now it says "the all"

packages/data-curation/src/data_curation/terminologies/utils/hl7.py
```diff
@@ -0,0 +1,131 @@
...
+            "description": "External Encounter",
+        }
+        hl7_rows.append(external_encounter)
+        # TODO: In Subsequent PR update this to be a logging statement
+        print(f"HL7 Encounter Act Codes Retrieved from HL7 Act Codes: {len(hl7_rows)}")
+        return hl7_rows
+
+
+def get_hl7_lab_interp() -> list[dict]:
+    """Function to get the all HL7 Codes and Terms for
```
bamader: Nit: as above, "all" and "the" are switched

packages/data-curation/src/data_curation/terminologies/hl7.py
```diff
@@ -0,0 +1,131 @@
...
+        # TODO: In Subsequent PR update this to be a logging statement
+        print(f"HL7 Lab Interpretation Record Count: {record_count}")
+
+        for hl7_row in hl7_codes:
+            hl7_code = hl7_row.get("code")
+            hl7_text = hl7_row.get("display")
+            # NOTE: we can add back in the definition as description, but there are some
+            # special character filtering we may need to do and some of the
+            # data in this field could clutter things up
+            # hl7_desc = hl7_row.get("definition")
```
bamader, line 120: Could you elaborate more on this (here on github, not necessarily in the code--I want to make sure I'm understanding the implications correctly)? Are we currently storing the `definition` field as our description? And what kinds of clutter are you seeing in the data?

bamader (approved):
Thanks for all your hard work on this Brady! This looks good to me since we've talked about the ways we want to handle embedding updates and those will be handled in a separate ticket. Just a few very minor documentation comments and one clarifying question for my own understanding.

packages/data-curation/src/data_curation/terminologies/utils/general.py
BradySkylight: I fixed this up and let me know if you find any places where I missed any of the updated paths @robertandremitchell

packages/data-curation/src/data_curation/terminologies/utils/general.py
BradySkylight: Yeah, I tried to import it but was having some issues doing so.  I can try again, but figured I would just get this working first.  I'll try again.

packages/data-curation/src/data_curation/terminologies/utils/hl7.py
BradySkylight: Nice.  Fixed!

packages/data-curation/src/data_curation/terminologies/hl7.py
BradySkylight: This is an old comment, but happy to follow-up with you later about this for sure.  I ran the API call again just to check and am familiar again with what was going on here.  Here is just one example of the 'descriptions' for one of the codes.  This is actually being pulled from a .json file in a FHIR ValueSet via the HL7 website.  It contains information that 'could be' and probably is used in HTML.

`"The observation/test result is interpreted as being outside the inclusion range for a particular protocol within which the result is being reported.\r\n\r\n**Example:** A positive result on a Hepatitis screening test.\r\n\r\n*Open Issue:* EX, HX, LX: These three concepts do not seem to meet a clear need in the vocabulary, and their use in observation interpretation appears likely to be covered by other existing concepts (e.g., A, H, L). The only apparent significant difference is their reference to use in protocols for exclusion of study subjects. These concepts/codes were proposed by RCRIM for use in the CTLaboratory message. They were submitted and approved in the November 2005 Harmonization cycle in proposal \"030103C\\_VOCAB\\_RCRIM\\_l\\_quade\\_RCRIM Obs Interp\\_20051028154455\". However, this proposal was not fully implemented in the vocabulary. The proposal recommended creation of the x\\_ClinicalResearchExclusion domain in ObservationInterpretation with a value set including those three concepts/codes, but there is no subdomain of that name or equivalent with a binding to either of the value sets that contain these concepts/codes. Members of the OO WG have recently attempted to contact members of RCRIM regarding these concepts, both by email and at the recent WGM in Atlanta, without response. It is felt by OO that the best course of action to take at this time is to add this comprehensive Open Issue rather than deprecate these three concepts at this time, until further discussion is held.",`

packages/data-curation/src/data_curation/terminologies/utils/general.py
BradySkylight: Got it to work!  But had to move things around again...but I like this structure better anyway.

packages/data-curation/src/data_curation/terminologies/loinc.py
```diff
@@ -0,0 +1,536 @@
...
+        :param loinc_code: The LOINC code that is either new or the existing one
+            where changes are required.
+        :param loinc_row: The actual LOINC data from the API for the LOINC code
+            being added/updated.
+        :param element_change: The list of changes required per LOINC code.
+
+        :returns: a list of embedding candidate records for the LOINC code.
+    """
+    emb_candidates = []
+    emb_candidate = {}
```
nickclyde: I think there's a bug here: `emb_candidate = {}` creates a single dict that gets mutated and appended multiple times below (lines ~510-534). Each `emb_candidates.append(emb_candidate)` adds the same object reference, so after the function returns, every entry in the list points to the same dict and reflects whichever `loinc_name_type`/`term` was assigned last. For a NEW LOINC, instead of getting 5 distinct candidate rows (short_name, long_name, display_name, full_name, consumer_name), the caller will get 5 references to one row.

This is because Python's `append()` function only does shallow copy / pass by reference. Explanation here: https://www.sobyte.net/post/2022-07/py-append/

Could we move the `emb_candidate = {...}` construction inside each `if` block (or use `emb_candidates.append({**emb_candidate, "loinc_name_type": ..., "term": ...})` to copy)? A small test covering NEW LOINC vs single-field change would also catch this going forward.

packages/data-curation/src/data_curation/terminologies/maintenance.py
```diff
@@ -0,0 +1,73 @@
...
+        # TODO: In Subsequent PR update this to be a logging statement
+        print(f"Getting all updates from LOINC since {loinc_version_date}!")
+        # get the current extract into a dict
+        loinc_current_dict = load_extract_file_to_dict(current_loinc_file)
+        loinc_updates = get_loinc_embedding_candidates(loinc_current_dict,loinc_version)
+    else:
+        # TODO: In Subsequent PR update this to be a logging statement
+        print(f"No updates found for the latest LOINC ({loinc_version}) Version!")
+    
+    if len(loinc_updates) > 0:
```
nickclyde: I believe `loinc_updates` can be unbound here. It's only assigned on line 30 inside the `if (file_date <= loinc_version_date):` branch; if the `else` branch runs (no update needed), `loinc_updates` is never defined and `len(loinc_updates) > 0` will raise `UnboundLocalError`.

Could we either initialize `loinc_updates = []` before the if/else, or move the `len(...) > 0` check inside the `if` branch?

packages/data-curation/src/data_curation/terminologies/general.py
```diff
@@ -0,0 +1,98 @@
...
+    return latest_file
```
nickclyde: If `files` is empty (e.g., the prefix doesn't match any file under `BASE_FOLDER`), `latest_file` is never assigned and the `return` raises `UnboundLocalError`. I think we'd want to handle the no-match case explicitly, maybe `return None` or raise a clear `FileNotFoundError` with the prefix in the message, so the maintenance flow surfaces a useful error if the LOINC extract is missing.

packages/data-curation/src/data_curation/terminologies/general.py
```diff
@@ -0,0 +1,98 @@
...
+
+        :param filename: Text of the filename to extact the date from.
+        :param terminology: The name of the value set in question as the
+            date formats may have different requirements based upon
+            the API response for said value set.
+
+        :returns: A formatted date string that can be used for comparison
+            to determine if an update is required.
+    """
+    file_date = re.search(r'\d{8}', filename).group()
```
nickclyde: Two notes here:

1. `re.search(r'\d{8}', filename).group()` will raise `AttributeError: 'NoneType' object has no attribute 'group'` if the filename doesn't contain an 8-digit run. A guard like `match = re.search(...); if match is None: raise ValueError(...)` would give a clearer error.
2. The function is annotated `-> str` but only returns inside the `if terminology == 'loinc':` branch; for any other terminology it falls through and returns `None`, which then gets compared as a date in `maintenance.py:25`. Could we raise on unknown terminology rather than silently returning `None`?

packages/data-curation/src/data_curation/terminologies/loinc.py
```diff
@@ -0,0 +1,536 @@
...
+        'Orders', or 'Both'.
+
+        :param version: Text string of the version number you want to 
+            use to filter LOINC codes for.
+
+        :returns: A list of dictionaries containing LOINC lab order records
+            including codes, terms, and axis information.
+    """
+    print(f"DIR: {LOINC_CS_NAMES}")
+    print(f"BASE: {BASE_FOLDER}")
```
nickclyde: These two prints (`DIR:` / `BASE:`) look like debug instrumentation left over from path debugging. Could we drop them?

packages/data-curation/src/data_curation/terminologies/loinc.py
```diff
@@ -0,0 +1,536 @@
...
+        :returns: boolean - true if it should be filtered and false if not
+    """
+    result: bool = False
+    for filter_text in LOINC_TEXT_TO_FILTER:
+        if filter_text in text:
+            result = True
+    return result
+
+
+def get_loinc_current_version_data() -> (str, str):
```
nickclyde: Nit: `-> (str, str)` isn't a valid type annotation (it's a tuple literal at runtime, not a tuple type); I think we want `-> tuple[str, str]`.

nickclyde (changes requested):
Awesome stuff Brady! This is a great step towards automating LOINC updates. A few thoughts:

- Tests: this PR adds a lotta lines! Even a small test module under packages/data-curation/tests/ might be worth considering!
- Shebang typo across new files: every new terminologies/*.py file starts with `#!/usr/bin/env python"""` (stray triple-quote on the shebang line). It happens to parse because # makes line 1 a comment, but it's a copy-paste artifact and worth cleaning to just `#!/usr/bin/env python` before the next round.
- Typos I noticed throughout: valusets (extraction docstring), multile (general.py docstring), dictiary (general.py docstring), Obsersavtions (loinc.py), defintion_desc (loinc.py local variable), contstruct (loinc.py docstring), hte (extraction comment).
- Print-based logging: I see the TODO comments calling these out for a follow-up PR, so leaving them alone, just noting that the volume of prints (especially debug-style ones like the DIR: / BASE: prints at the top of get_loinc_lab_orders) makes the followup more valuable.

Thanks for the work! The comments below focus on a few bugs I found.

packages/data-curation/src/data_curation/terminologies/loinc.py
BradySkylight: Nice find.  Doah!  Fixed.

packages/data-curation/src/data_curation/terminologies/loinc.py
BradySkylight: Yes, those should have been removed.  Fixed.

packages/data-curation/src/data_curation/terminologies/loinc.py
BradySkylight: This is a good catch.  This is where I trip up in python.  Will figure a way to resolve this.  You can see what I'm trying to do though.  The loinc code, type, and axis data are the same for multiple 'entries' of embeddings but the term used, from said code, is the difference in the record, but each need to be added.

packages/data-curation/src/data_curation/terminologies/utils/general.py
robertandremitchell: lgtm! thanks brady!

packages/data-curation/src/data_curation/terminologies/loinc.py
BradySkylight: I have addressed this in my latest commit/push... adding tests to verify!

packages/data-curation/src/data_curation/terminologies/maintenance.py
BradySkylight: This was just a silly mistake on my part.  Fixed!

packages/data-curation/src/data_curation/terminologies/general.py
BradySkylight: Great suggestion.  Fixed

packages/data-curation/src/data_curation/terminologies/general.py
BradySkylight: great catch!  Fixed.

nickclyde (approved):
Amazing work Brady! Thanks for adding tests, LGTM!!

PR #536, opened 2026-05-07
Creating deterministic augmentation IDs

### LINKED ISSUE #535: Update Augmentation - Deterministic Ids

robertandremitchell:
## Description

Changes eicr_augmentation.py to have deterministic IDs, updates subsequent code and tests.

## Related Issues

Closes #535 

## Additional Notes

## Checklist

## Checklist for Reviewers


BradySkylight:
As part of this PR we should also just grab the versionNumber of the incoming eicr and use that as our 'augmented' eicr version as well.

robertandremitchell:
> As part of this PR we should also just grab the versionNumber of the incoming eicr and use that as our 'augmented' eicr version as well.

thanks, @BradySkylight ! Should be fixed now, but curious if you think our fallback should be hardcoding to versionNumber=1 versus raising an error? Right now I added a ValueError, but idk if we expect that this would actually trip up eICRs

BradySkylight:
> > As part of this PR we should also just grab the versionNumber of the incoming eicr and use that as our 'augmented' eicr version as well.
> 
> thanks, @BradySkylight ! Should be fixed now, but curious if you think our fallback should be hardcoding to versionNumber=1 versus raising an error? Right now I added a ValueError, but idk if we expect that this would actually trip up eICRs

I think defaulting to 1 if there isn't a version number present makes sense.

robertandremitchell:
> > > As part of this PR we should also just grab the versionNumber of the incoming eicr and use that as our 'augmented' eicr version as well.
> > 
> > 
> > thanks, @BradySkylight ! Should be fixed now, but curious if you think our fallback should be hardcoding to versionNumber=1 versus raising an error? Right now I added a ValueError, but idk if we expect that this would actually trip up eICRs
> 
> I think defaulting to 1 if there isn't a version number present makes sense.

makes sense, updated!

packages/augmentation/src/augmentation/services/eicr_augmenter.py
```diff
@@ -197,10 +209,12 @@ def _get_new_effective_time(self) -> Element:
...
         """Generate a versionNumber element for the augmented eICR document."""
+        original_version_number = self._get_original_by_xpath("/ClinicalDocument/versionNumber")
+        original_version_number_value = (
+            original_version_number.get("value") or "1"
+        )  # default to "1" if value attribute is missing to avoid errors
         version_number_tag = etree.Element("versionNumber")
-        # hard code to 1 for now
-        # TODO: we may need to have some way to increment this later
-        version_number_tag.set("value", "1")
+        version_number_tag.set("value", original_version_number_value)
```
nickclyde: `_get_old_version_number()` already does `self._get_original_by_xpath("/ClinicalDocument/versionNumber")` (line 179), so we could call that here to avoid the duplication.

packages/augmentation/tests/unit/test_eicr_augmenter.py
```diff
@@ -157,19 +148,48 @@ def test_get_old_document_id_preserves_assigning_authority_name_when_present(sel
...
+
+        assert version_number.get("value") == "1"
+
+    def test_generates_same_augmented_ids_for_same_seed(self):
+        """Tests deterministic augmented identifiers use stable values for the same seed."""
+        first_augmenter = EICRAugmenter(BASIC_ECR, [])
+        second_augmenter = EICRAugmenter(BASIC_ECR, [])
+
+        assert first_augmenter.new_doc_id == second_augmenter.new_doc_id
+        assert first_augmenter.new_set_id == second_augmenter.new_set_id
```
nickclyde: I think the test name might be a bit misleading, since neither augmenter is given a seed, this is really testing "no-seed callers default to the same value when the XML is the same." To actually test "same seed produces same IDs", could we pass an explicit `deterministic_id_seed` (and maybe even use two different XML payloads with the same seed, to prove the seed, not the document, is what determines the ID)?

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -124,6 +124,7 @@ def _process_record(record: SQSRecord, s3_client: BaseClient) -> None:
             document=original_eicr,
             nonstandard_codes=augmenter_input.nonstandard_codes,
             config=config,
+            deterministic_id_seed=augmenter_input.eicr_id,
```
nickclyde: Just to confirm my understanding: `augmenter_input.eicr_id` here is the S3 persistence_id (e.g. `2025/09/03/1-5f84c7a5-...`), not the eICR's `<id root>` UUID, right?

nickclyde (commented):
Nice work Rob!!

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
robertandremitchell: yeah, the augmenter_input.eicr_id is the S3 persistence_id parsed from the TTC output object key, not the eICR <id root>. My thought was the persistence_id as the deterministic seed would ensure IDs are stable.

packages/augmentation/tests/unit/test_eicr_augmenter.py
robertandremitchell: yeah, that makes sense. I've renamed it for clarity and added the additional test tha ttests the basic vs. covid eicr with the deterministic ID.

packages/augmentation/src/augmentation/services/eicr_augmenter.py
robertandremitchell: ack, good call. updated!

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
JNygaard-Skylight: Doesn't have to be in this PR, but that attribute should be renamed as in every other context eICR ID would mean the value of the ID element.

robertandremitchell: totally makes sense. updated!

nickclyde (approved):
Great work Rob!

PR #543, opened 2026-05-11
Preserve namespaces during augmentation

nickclyde:
## Description

Fixes a bug reported by Geo where every augmented eICR submitted to RCKMS in dev is rejected with `responseCode 422 / "Payload is missing or empty"`.

### The bug

The augmenter parses incoming eICR XML with `clean_xml_tree()`, which strips every namespace declaration off the tree so XPath queries can be written without prefixes (e.g. `/ClinicalDocument/id` instead of `/cda:ClinicalDocument/cda:id`). On the way out, the tree is re-serialized with `etree.tostring()`, but nothing puts the namespaces back. The result:

```xml
<!-- input -->
<ClinicalDocument xmlns="urn:hl7-org:v3" xmlns:cda="urn:hl7-org:v3"
  xmlns:sdtc="urn:hl7-org:sdtc" xmlns:voc="http://www.lantanagroup.com/voc"
  xmlns:xsi="..." xsi:schemaLocation="...">

<!-- output before this PR -->
<ClinicalDocument xmlns:xsi="..." xsi:schemaLocation="...">
```

The default namespace `xmlns="urn:hl7-org:v3"` is gone, which means every child element (`<id>`, `<setId>`, the whole CDA body) parses into the null namespace. RCKMS doesn't recognize the payload as a CDA `ClinicalDocument` and 422s the whole batch.

This was blocking the training pipeline; every augmented eICR was failing routing.

### The fix

Stop stripping namespaces. Parse the eICR with namespaces intact and make the augmenter namespace-aware:

- `parse_eicr_xml()` replaces `clean_xml_tree()`. It just parses, no stripping.
- `_get_element_by_xpath()` rewrites incoming unprefixed XPaths to use a `cda:` prefix bound to `urn:hl7-org:v3`, so existing callers (including `schematron_error_xpath` values supplied by `NonstandardCodeInstance`) keep working unchanged.
- A `_cda_element()` factory creates new elements (`<id>`, `<setId>`, `<author>`, `<translation>`, etc.) in the CDA default namespace, so they inherit the namespace and don't trigger `xmlns=""` escapes on serialization.
- `_absolute_local_xpath()` replaces `getroottree().getpath()` for producing the `new_translation_xpath` metadata field, because lxml's `getpath()` falls back to ambiguous `/*[N]` wildcards under default namespaces.

## How `cda_xpath` works

`cda_xpath()` (in `eicr_utils.py`) is a small XPath rewriter. It takes an unprefixed XPath that targets the eICR and prepends `cda:` to every element step that isn't already prefixed, isn't an attribute, isn't a wildcard, and isn't a function call. The augmenter then evaluates the rewritten XPath with `namespaces={"cda": "urn:hl7-org:v3"}`, so it resolves correctly against a namespace-bearing tree.

It works on the path *string* (no parsing of the tree) by splitting on `/` and rewriting each segment.

### Walkthrough

```python
def cda_xpath(xpath: str) -> str:
    parts = xpath.strip().split("/")
    rewritten = []
    for part in parts:
        if not part or part.startswith("@") or part == "*" or ":" in part or "(" in part:
            rewritten.append(part)
        else:
            rewritten.append(f"cda:{part}")
    return "/".join(rewritten)
```

Splitting on `/` leaves empty strings at the boundaries; that's intentional. `"/ClinicalDocument/id".split("/")` is `["", "ClinicalDocument", "id"]`, and `"//id".split("/")` is `["", "", "id"]`. The empty strings are preserved so the join puts the slashes back where they belong, including the descendant-or-self `//` form.

Each non-empty segment is left alone if it's already prefixed (`cda:foo`), an attribute step (`@root`), a bare wildcard (`*`), or contains a function call (`text()`, `local-name()`, `position()`, anything with `(`). Everything else is assumed to be a CDA element step and gets `cda:` prepended. Not sure if this is the right call, I'm not super well-versed in how CDA works, but I figured it was a sensible default.

The leading `.strip()` is there because schematron `<context>` elements emitted upstream often carry surrounding indentation whitespace. lxml's `xpath()` seems to allow this, but our naive split-on-`/` rewriter won't (a leading `"\n    "` becomes its own segment and gets `cda:` prepended, producing an invalid expression). This was actually caught by the e2e suite during development!

### Examples

| Input | Output |
|---|---|
| `/ClinicalDocument` | `/cda:ClinicalDocument` |
| `/ClinicalDocument/id` | `/cda:ClinicalDocument/cda:id` |
| `/ClinicalDocument/id/@root` | `/cda:ClinicalDocument/cda:id/@root` |
| `/ClinicalDocument/component/structuredBody/component/section/entry/component/observation` | `/cda:ClinicalDocument/cda:component/cda:structuredBody/cda:component/cda:section/cda:entry/cda:component/cda:observation` |
| `//id` | `//cda:id` (empty middle segment preserved) |
| `/cda:ClinicalDocument/cda:id` | `/cda:ClinicalDocument/cda:id` (already prefixed, left alone) |
| `/ClinicalDocument/text()` | `/cda:ClinicalDocument/text()` (function call, left alone) |
| `"\n    /ClinicalDocument/component[1]/observation[1]  "` | `/cda:ClinicalDocument/cda:component[1]/cda:observation[1]` (stripped, then rewritten) |

### Scope and limits

`cda_xpath()` deliberately handles only the simple-path forms that the augmenter actually uses today: absolute paths, attribute steps, the descendant-or-self abbreviation, and bare function calls at the end. It does **not** parse XPath predicates. `entry[code/@code='123']` will get the outer `entry` prefixed, but the `code/@code='123'` inside the brackets stays untouched - wasn't sure if this is wrong, would it need it prefixed too? AFAICT, none of the current callers use predicates.

## How `_absolute_local_xpath` works

The augmenter writes the path of every new `<translation>` element it injects into the metadata's `new_translation_xpath` field, so downstream consumers can locate the augmented element. Before this PR, that path came from `self._augmented_element.getroottree().getpath(new_translation)`.

`getpath()` is the obvious choice (it's lxml's built-in "give me an absolute XPath to this element"), but it has a quirk under default namespaces. From what I read, when an element belongs to a namespace and lxml can't find an unambiguously-bound prefix for it in scope, `getpath()` falls back to numbered wildcards. For our augmented tree, where elements sit in `urn:hl7-org:v3` (declared as both the default namespace *and* under the `cda` prefix), `getpath()` returns paths like:

```
/*/*[6]/*/*/*/*/*/*/*[1]/*[2]
```

That's not useful and jumbled, and it's a regression vs. the pre-fix snapshot which had clean paths like `/ClinicalDocument/.../translation`. So `_absolute_local_xpath()` builds the path by walking the element's ancestry directly:

```python
def _absolute_local_xpath(element: Element) -> str:
    parts: list[str] = []
    current: Element | None = element
    while current is not None:
        local = etree.QName(current).localname
        parent = current.getparent()
        if parent is not None:
            same_tag_siblings = [
                c
                for c in parent
                if not isinstance(c, etree._Comment) and etree.QName(c).localname == local
            ]
            if len(same_tag_siblings) > 1:
                idx = same_tag_siblings.index(current) + 1
                parts.append(f"{local}[{idx}]")
            else:
                parts.append(local)
        else:
            parts.append(local)
        current = parent
    return "/" + "/".join(reversed(parts))
```

### How the walk produces a path

1. Start at the target element.
2. At each step, take the element's **local name** (`etree.QName(current).localname` strips any namespace, so `{urn:hl7-org:v3}translation` becomes `translation`). This is what makes the output unprefixed.
3. Look at the parent's children that share that same local name. If the element is the only one with that name among its siblings, emit just the tag. If there are several, emit `tag[N]` with a 1-based index, matching XPath's positional-predicate semantics. Comments are skipped so they never shift the indexing.
4. Walk up to the parent and repeat. Stop when `getparent()` returns `None`, which is the root.
5. The parts were collected from leaf to root, so reverse and join with `/`, prepending a leading `/` for an absolute path.

### Examples

Given the augmented `<translation>` element from `test_basic_eicr` (in a tree where each element is the only one of its kind among its siblings), the walk goes:

```
translation -> code -> observation -> component -> entry -> section
              -> component -> structuredBody -> component -> ClinicalDocument
```

Each step finds zero same-tag siblings, so no indices are emitted. Result:

```
/ClinicalDocument/component/structuredBody/component/section/entry/component/observation/code/translation
```

If the eICR had multiple `<entry>` elements under the same `<section>` (which I think would be the case when there are multiple lab results under one section), the helper would emit `entry[2]` for the second one:

```
/ClinicalDocument/component/structuredBody/component/section/entry[2]/component/observation/code/translation
```

A few things to note:

- **Comments are filtered before indexing.** lxml exposes `<!-- ... -->` nodes alongside element children. The augmenter inserts a lot of these (`<!--DATA AUGMENTATION: ... -->`), and if you counted them in `same_tag_siblings` you'd get incorrect indices that depend on whether comments were inserted before or after the element. The `isinstance(c, etree._Comment)` filter keeps the indexing element-only.
- **Indices are emitted only when needed.** If `<observation>` is unique among its parent's children, you get `observation`, not `observation[1]`. This matches lxml's behavior on the namespace-stripped tree and keeps the metadata readable in the common case.

## Regression test

Geo provided the original eICR he used to surface this bug (the one that RCKMS rejected in dev). It's now wired into the local e2e suite as a regression test:

- **Fixtures**: `e2e/assets/namespace_preservation_eicr.xml` (Geo's eICR) and `e2e/assets/namespace_preservation_schematron_errors.xml` (the schematron output that drove the TTC pipeline against it).
- **Test**: `TestNamespacePreservation::test_augmented_eicr_preserves_cda_namespaces` in `e2e/test_e2e.py`. Runs the full TTC + augmentation pipeline locally against Geo's fixtures and asserts:
  1. `augmented_root.nsmap == input_nsmap`: every namespace declaration from the input (`xmlns="urn:hl7-org:v3"`, `cda`, `sdtc`, `voc`, `xsi`) is present on the output. Before the fix the output nsmap was `{"xsi": "..."}` only.
  2. `augmented_root.tag == "{urn:hl7-org:v3}ClinicalDocument"`: the root is in the CDA namespace, not the null namespace. Before the fix the tag was bare `"ClinicalDocument"`.
  3. Augmenter-injected `<translation>` elements resolve via `//cda:translation` xpath and sit in `urn:hl7-org:v3`. Before the fix this xpath would have returned an empty list.
No issue logged

## Additional Notes

- **`text-to-code` package was intentionally left alone.** `packages/text-to-code/src/text_to_code/services/eicr_processor.py::_create_xml_tree` has the same namespace-stripping pattern, but it's only used for read-only candidate extraction (no serialization), so it won't cause RCKMS rejection.
- **Snapshot updates.** Four snapshot files were regenerated. The only diff vs. the previous snapshots is the four namespace declarations now present on the root `<ClinicalDocument>`. Child elements remain unprefixed and inherit the default namespace; no `xmlns=""` escapes.
- **Verified end-to-end locally** by parsing the regenerated snapshot with lxml and confirming the root resolves as `{urn:hl7-org:v3}ClinicalDocument`, the original namespace map is intact (`None`, `cda`, `sdtc`, `voc`, `xsi`), and an injected `<translation>` element sits in `urn:hl7-org:v3`. Let's ask Geo to re-run a submission through the augmenter lambda in dev once this ships, to confirm RCKMS no longer 422s.


robertandremitchell (approved):
makes sense to me. awesome add for the e2e suite, too!

BradySkylight (approved):
Great catch and fix and plenty of details for documentation!

Issue #533, opened 2026-05-06
Cache TTC pipeline results in a second OpenSearch index


nickclyde:
### Summary

Add a result cache to the TTC Lambda pipeline so that repeat text candidates skip embedding, KNN, and reranking. Cache lives in a new OpenSearch index in the existing cluster, keyed by `(normalized_text, data_field)` with the final reranked LOINC `Code` as the value.

### Motivation

Common lab test names (e.g. "Hemoglobin A1c", "BUN", "Creatinine") will likely repeat across eICRs. Currently every pipeline run pays for a SentenceTransformer encode, an OpenSearch KNN query, and a CrossEncoder rerank. The pipeline output is deterministic on `(candidate text, data field)`, so a content-addressable cache eliminates the redundant work without changing semantics.

### Proposed design

**New cache index** (`ttc-result-cache`, same OpenSearch domain, no KNN field):

```json
{
  "mappings": {
    "properties": {
      "cache_key":   {"type": "keyword"},
      "text":        {"type": "keyword"},
      "data_field":  {"type": "keyword"},
      "code":        {"type": "object", "enabled": false},
      "score":       {"type": "float"},
      "cached_at":   {"type": "date"}
    }
  }
}
```

**Key**: `sha256(text.strip().lower() + "|" + data_field.value)`, used as the OpenSearch doc `_id`.

**Value**: serialized `Code` (from `shared_models`) plus the reranker score and a `cached_at` timestamp for diagnostics.

**Pipeline integration**: in `text-to-code-lambda/lambda_function.py`, between `select_relevant_text` and `embed`:
1. Compute key, attempt `GET /ttc-result-cache/_doc/{key}`.
2. Hit → reconstruct `Code`, build `NonstandardCodeInstance`, emit `cache_hit` metric, continue.
3. Miss → run existing embed/query/rerank, write the result back to the cache index, emit `cache_miss` metric.

Encapsulate this in a new `ResultCache` service under `packages/text-to-code/src/text_to_code/services/result_cache.py`.

**Invalidation**: wipe-and-refill on reindex. The `ttc-index-lambda` is invoked at deploy time; extend it to `DELETE /ttc-result-cache` and recreate the empty index. When the OSIS pipeline reloads LOINC data, a follow-up invocation of the index Lambda clears the cache so stale entries can't survive a LOINC refresh.

### Tasks

[...]

PR #546, opened 2026-05-12
Add TF Config and Index Lambda Extension

bamader:
## Description

This PR manages the terraform and the index lambda portion of creating a result cache for previously seen eICR inputs. In terraform, this change set includes:

* Adding a new environment variable for the result cache
* Creating a bootstrap function for the result cache index to initialize
* Adding the result cache bootstrap as a dependency to the pipeline
* Updating documentation (as well as spikes which reference the index procedures)

In the Index Lambda, this PR extends the existing functionality to also apply to the Result Cache index by defining some new actions (basically just counterparts of the Vector Search Index) and re-using functions there. Unit tests are updated and new tests are added accordingly.

Finally, our Index Lambda previously had a set of kruft code which was a carryover from our very first embedding trial. The block of code dealing with checking whether the OpenSearch Index had knn vectors in its property mappings (in the create function) is now no longer necessary: (1) the result cache index does not use KNN embeddings, and (2) (more importantly) the only reason this code was originally there is because our first AWS index upload back in September of 2025 did not have this property set correctly for some of its vectors. When we went to create the index, we had to do an additional check to see whether the index first needed to be deleted so things could be freshly set. Now, though, all of our vectors have the same uniform properties set, so this is a non-issue and will not be a problem going forward. The unit test dealing with this behavior has also been deleted.

## Related Issues

Partially addresses some tasks in #533 .

### DIFF
--- a/packages/index-lambda/src/index_lambda/lambda_function.py
+++ b/packages/index-lambda/src/index_lambda/lambda_function.py
@@ -16,8 +16,8 @@ class OpenSearchIndexMapping(TypedDict):
     data to-index), but these attributes are themselves required.
     """
 
-    settings: str
-    mappings: str
+    settings: dict
+    mappings: dict
 
 
 logger = Logger(service="index-lambda")
--- a/packages/index-lambda/src/index_lambda/lambda_function.py
+++ b/packages/index-lambda/src/index_lambda/lambda_function.py
@@ -36,16 +36,36 @@ INDEX_MAPPING = {
     },
 }
 
+RESULT_CACHE_INDEX_MAPPING = {
+    "settings": {"index": {"number_of_shards": 1, "number_of_replicas": 1, "knn": False}},
+    "mappings": {
+        "properties": {
+            "cache_key": {"type": "keyword"},
+            "text": {"type": "keyword"},
+            "data_field": {"type": "keyword"},
+            "code": {"type": "object", "enabled": False},
+            "score": {"type": "float"},
+            "cached_at": {"type": "date"},
+        },
+    },
+}
+
 
 @logger.inject_lambda_context
 def handler(event: dict, context: LambdaContext) -> dict:
     """Lambda function to manage the OpenSearch index for LOINC code embeddings.
 
-    Supports two actions via the event dict:
+    Supports six actions via the event dict:
     - "clear_index": Deletes the existing index (if any) and recreates it empty.
       Use this before re-ingesting embeddings to avoid duplicates.
+    - "clear_result_cache": As above, but for the Result Cache index rather than
+      the Vector Search index.
     - "create_index" (default): Creates the index only if it doesn't exist,
       and self-heals incorrect mappings.
+    - "create_result_cache": As above, but for the Result Cache index.
+    - "set_slowlog": Changes logging parameters for AWS across multiple types
+      of logging information.
+    - "set_result_cache_slowlog": As above, but for the Result Cache index.
 
     :param event: The event dict passed by AWS Lambda. Reads "action" key.
     :param context: The context dict passed by AWS Lambda (not used).
@@ -53,42 +73,59 @@ def handler(event: dict, context: LambdaContext) -> dict:
     aws_auth = lambda_handler.create_aws_auth()
     os_client = lambda_handler.create_opensearch_client(aws_auth)
     index_name = get_env_variable("INDEX_NAME")
+    result_cache_index_name = get_env_variable("RESULT_CACHE_INDEX_NAME")
 
     action = event.get("action", "create_index") if event else "create_index"
-
+    logger_name = (
+        index_name
+        if action in ["create_index", "set_slowlog", "clear_index"]
+        else result_cache_index_name
+    )
     with logger.append_context_keys(
-        index_name=index_name,
+        index_name=logger_name,
         action=action,
     ):
         logger.info("Index Lambda started", status="processing")
 
         if action == "clear_index":
-            result = _clear_index(os_client, index_name)
+            result = _clear_index(os_client, index_name, INDEX_MAPPING, action)
+        elif action == "clear_result_cache":
+            result = _clear_index(
+                os_client, result_cache_index_name, RESULT_CACHE_INDEX_MAPPING, action
+            )
         elif action == "set_slowlog":
             result = _set_slowlog(os_client, index_name, event.get("threshold_ms", 0))
+        elif action == "set_result_cache_slowlog":
+            result = _set_slowlog(os_client, result_cache_index_name, event.get("threshold_ms", 0))
+        elif action == "create_index":
+            result = _create_index(os_client, index_name, INDEX_MAPPING)
         else:
-            result = _create_index(os_client, index_name)
+            result = _create_index(os_client, result_cache_index_name, RESULT_CACHE_INDEX_MAPPING)
 
         logger.info("Index Lambda completed", status="success")
 
         return result
 
 
-def _clear_index(os_client: OpenSearch, index_name: str) -> dict:
-    """Delete the index if it exists, then recreate it with correct mappings.
+def _clear_index(os_client: OpenSearch, index_name: str, index_mapping: dict, action: str) -> dict:
+    """Delete the specified index if it exists, then recreate it with the supplied mapping.
 
     :param os_client: The OpenSearch client
     :param index_name: The name of the index
+    :param index_mapping: The formally-specified mapping to create for this index.
+    :param action: The action from the event queue that kicked off the index clear.
+      We use this to specify appropriate logging information in the response.
     """
     deleted = False
     if os_client.indices.exists(index=index_name):
         os_client.indices.delete(index=index_name)
         deleted = True
 
-    os_client.indices.create(index=index_name, body=INDEX_MAPPING)
+    os_client.indices.create(index=index_name, body=index_mapping)
 
     logger.info(
         "OpenSearch index cleared",
+        index_name=index_name,
         index_deleted=deleted,
         index_recreated=True,
         status="success",
@@ -96,7 +133,8 @@ def _clear_index(os_client: OpenSearch, index_name: str) -> dict:
 
     return {
         "statusCode": 200,
-        "action": "clear_index",
+        "action": action,
+        "index_name": index_name,
         "index_deleted": deleted,
         "index_recreated": True,
     }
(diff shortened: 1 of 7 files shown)


packages/index-lambda/src/index_lambda/lambda_function.py
```diff
@@ -36,67 +36,105 @@
...
+            result = _create_index(os_client, result_cache_index_name, RESULT_CACHE_INDEX_MAPPING)
 
         logger.info("Index Lambda completed", status="success")
 
         return result
 
 
-def _clear_index(os_client: OpenSearch, index_name: str) -> dict:
-    """Delete the index if it exists, then recreate it with correct mappings.
+def _clear_index(os_client: OpenSearch, index_name: str, index_mapping: dict, action: str) -> dict:
```
JNygaard-Skylight: Could there be more specific typing, such as a `typeddict`, or `dataclass` for `index_mapping`?

packages/index-lambda/src/index_lambda/lambda_function.py
```diff
@@ -127,46 +165,42 @@ def _set_slowlog(os_client: OpenSearch, index_name: str, threshold_ms: int) -> d
...
         "index": index,
+        "index_name": index_name,
         "settings": settings,
         "mappings": mappings,
     }
 
 
-def _create_index(os_client: OpenSearch, index_name: str) -> dict:
-    """Create the index if it doesn't exist, self-healing incorrect mappings.
+def _create_index(os_client: OpenSearch, index_name: str, index_mapping: dict) -> dict:
```
JNygaard-Skylight: Could there be more specific typing, such as a `typeddict`, or `dataclass` for `index_mapping`?

terraform/README.md
```diff
@@ -67,9 +67,11 @@ Each Lambda function has its own IAM role scoped to least-privilege S3 permissio
...
+Deployed as a **container image** from ECR (`package_type = "Image"`) using `Dockerfile.index` at repo root. Responsible for creating the OpenSearch KNN Index and the OpenSearch Result Cache Index at deploy time. It is **invoked by Terraform** (`aws_lambda_invocation.index_bootstrap`) during `terraform apply`, before the ingestion pipeline is created.
 
-The index it creates has LOINC-specific field mappings including `description_vector` (1024-dimension `knn_vector` using HNSW/faiss/cosine), `loinc_type`, `loinc_code`, `loinc_name_type`, and other LOINC metadata fields. Uses the `lambda_handler` shared utilities and reads `OPENSEARCH_ENDPOINT_URL` from its environment.
+The first Index it creates--the Vector Search Index--has LOINC-specific field mappings including `description_vector` (1024-dimension `knn_vector` using HNSW/faiss/cosine), `loinc_type`, `loinc_code`, `loinc_name_type`, and other LOINC metadata fields. Uses the `lambda_handler` shared utilities and reads `OPENSEARCH_ENDPOINT_URL` from its environment.
```
JNygaard-Skylight: Is the `--` around the `Vector Search Index` meant to be formatting? As in `_Vector Search Index_`?

terraform/README.md
```diff
@@ -67,9 +67,11 @@ Each Lambda function has its own IAM role scoped to least-privilege S3 permissio
...
+
+The second Index it creates--the Result Cache Index--contains the hashed results of previously computed embeddings and nearest neighbor queries, so that when the pipeline later receives eICRs containing a previously hashed value, the correct standardized code can simply be looked-up, rather than re-embedded and re-ranked. The Result Cache Index shares handling with the `lambda_handler` using the same functions but different actions than the Vector Search Index.
```
JNygaard-Skylight: Same question as above regarding the double hyphens.

packages/index-lambda/src/index_lambda/lambda_function.py
```diff
@@ -36,67 +36,105 @@
...
+            )
         elif action == "set_slowlog":
             result = _set_slowlog(os_client, index_name, event.get("threshold_ms", 0))
+        elif action == "set_result_cache_slowlog":
+            result = _set_slowlog(os_client, result_cache_index_name, event.get("threshold_ms", 0))
+        elif action == "create_index":
+            result = _create_index(os_client, index_name, INDEX_MAPPING)
         else:
-            result = _create_index(os_client, index_name)
+            result = _create_index(os_client, result_cache_index_name, RESULT_CACHE_INDEX_MAPPING)
```
nickclyde, line 116: I think this `else:` branch is a bit risky as written — any action string that doesn't match the earlier `elif`s (typos, stale payloads, or a future action that hasn't been wired up yet) will silently run `_create_index` against the result-cache index. The previous version defaulted unknown actions to `create_index` (main index), so this is a behavior change too.

Could we make the result-cache branch explicit and let unknown actions raise? Something like:

```python
elif action == "create_result_cache":
    result = _create_index(os_client, result_cache_index_name, RESULT_CACHE_INDEX_MAPPING)
else:
    raise ValueError(f"Unknown action: {action!r}")
```

That way the docstring's enumerated list of six actions is accurate.

packages/index-lambda/src/index_lambda/lambda_function.py
```diff
@@ -36,67 +36,105 @@
...
     index_name = get_env_variable("INDEX_NAME")
+    result_cache_index_name = get_env_variable("RESULT_CACHE_INDEX_NAME")
 
     action = event.get("action", "create_index") if event else "create_index"
-
+    logger_name = (
+        index_name
+        if action in ["create_index", "set_slowlog", "clear_index"]
+        else result_cache_index_name
+    )
```
nickclyde: Small nit — the variable is called `logger_name` but it's really the index name we're operating on, used both for log context and (indirectly via the dispatch below) which index gets touched. Also, the allow-list of action strings here has to stay in sync with the dispatch below; if a new action is added and forgotten here, logs would attribute it to the wrong index.

Would it be cleaner to compute it from the action itself, e.g. `result_cache_index_name if "result_cache" in action else index_name`, or to just attach `index_name=index_name` (or the resolved one) inside each branch where we already know which index applies?

packages/index-lambda/src/index_lambda/lambda_function.py
```diff
@@ -127,46 +165,42 @@ def _set_slowlog(os_client: OpenSearch, index_name: str, threshold_ms: int) -> d
...
+        "index_name": index_name,
         "settings": settings,
         "mappings": mappings,
     }
 
 
-def _create_index(os_client: OpenSearch, index_name: str) -> dict:
-    """Create the index if it doesn't exist, self-healing incorrect mappings.
+def _create_index(os_client: OpenSearch, index_name: str, index_mapping: dict) -> dict:
+    """Create the specified index if it doesn't exist, self-healing incorrect mappings.
```
nickclyde: Docstring still says "self-healing incorrect mappings" but the self-heal block was removed (intentionally, per the PR description). Maybe rewrite to "Create the index if it doesn't exist; otherwise return the current settings and mappings."

packages/index-lambda/src/index_lambda/lambda_function.py
```diff
@@ -127,46 +165,42 @@ def _set_slowlog(os_client: OpenSearch, index_name: str, threshold_ms: int) -> d
...
+    :param os_client: The OpenSearch client.
+    :param index_name: The name of the index.
+    :param index_mapping: The properties mapping for the specified index.
     """
+    created = False
     if not os_client.indices.exists(index=index_name):
-        os_client.indices.create(index=index_name, body=INDEX_MAPPING)
-        logger.info("OpenSearch index created", status="success")
+        os_client.indices.create(index=index_name, body=index_mapping)
+        logger.info(f"OpenSearch index {index_name} created", status="success")
```
nickclyde: Tiny consistency thing — elsewhere in this file we pass `index_name` as a structured field on `logger.info(...)` rather than interpolating it into the message. Could we do the same here so the message is constant across invocations?

```python
logger.info("OpenSearch index created", index_name=index_name, status="success")
```

packages/index-lambda/tests/test_index_lambda_function.py
```diff
@@ -1,73 +1,118 @@
...
-        self.create_calls: list[str] = []
-        self.put_settings_calls: list[tuple[str, dict]] = []
+        self._exists_calls = {INDEX_NAME: 0, RESULT_CACHE_INDEX_NAME: 0}
+        self._initially_exists = {
+            INDEX_NAME: index_initially_exists,
+            RESULT_CACHE_INDEX_NAME: result_cache_initially_exists,
+        }
+        self.delete_calls: dict[list[str]] = {INDEX_NAME: [], RESULT_CACHE_INDEX_NAME: []}
+        self.create_calls: dict[list[str]] = {INDEX_NAME: [], RESULT_CACHE_INDEX_NAME: []}
+        self.put_settings_calls: dict[list[tuple[str, dict]]] = {
```
nickclyde: I believe these annotations should be `dict[str, list[...]]` rather than `dict[list[...]]`. `dict` takes two parameters (key type, value type), so the current form is silently malformed. I think it happens to evaluate at runtime as a `GenericAlias` and `ty` doesn't seem to be complaining, but as a type description it's misleading.

```python
self.delete_calls: dict[str, list[str]] = {...}
self.create_calls: dict[str, list[str]] = {...}
self.put_settings_calls: dict[str, list[tuple[str, dict]]] = {...}
```

packages/index-lambda/tests/test_index_lambda_function.py
```diff
@@ -112,55 +160,88 @@ def test_handler_success(self, monkeypatch, mock_lambda_context):
             }
         }
 
-    def test_handler_recreates_index_when_vector_mapping_incorrect(
-        self, monkeypatch, mock_lambda_context
-    ):
-        """Test handler recreates the index when description_vector mapping is not knn_vector."""
-        patch_lambda_handler(monkeypatch, "keyword")
+    def test_handler_create_result_cache_success(self, monkeypatch, mock_lambda_context):
+        """Test handler creates the result cache index when it already exists."""
```
nickclyde: Small nit: this test actually exercises the path where the result-cache index *already exists* and `_create_index` no-ops (`ran_index_creation: False`). The docstring reads like a creation-path test. Maybe something like "Test handler returns existing metadata when the result cache index already exists."

packages/index-lambda/tests/test_index_lambda_function.py
```diff
@@ -179,14 +260,17 @@ def test_handler_set_slowlog_updates_index_settings(self, monkeypatch, mock_lamb
...
+                    {
+                        "index.search.slowlog.threshold.query.warn": f"{THRESHOLD_MS}ms",
+                        "index.search.slowlog.threshold.query.info": f"{THRESHOLD_MS}ms",
+                        "index.search.slowlog.threshold.fetch.warn": f"{THRESHOLD_MS}ms",
+                        "index.search.slowlog.threshold.fetch.info": f"{THRESHOLD_MS}ms",
+                    },
+                )
+            ],
+            RESULT_CACHE_INDEX_NAME: [],
+        }
```
nickclyde, line 285: Could we add a sibling test for `set_result_cache_slowlog`? a parallel of `test_handler_set_slowlog_updates_index_settings` would close the coverage gap.

nickclyde (commented):
Nice work Brandon! Re-using _clear_index and _create_index with a mapping/name parameter keeps the new actions small. A few things notes below:

packages/index-lambda/src/index_lambda/lambda_function.py
bamader: I can add that, yeah. We won't know the specifications for each of the properties we're requiring (`settings` and `mappings`) since they're dependent on what data is being stored, but I can at least enumerate that those fields are themselves required.

terraform/README.md
bamader: No, that was an emdash. I'm used to entering them in word processors or auto-formatting markdown where double hyphen autocorrects to a single emdash. I've replaced them with the actual character so it renders properly.

packages/index-lambda/src/index_lambda/lambda_function.py
bamader: Yeah this is a great call. Previously we had an `else` block here that just ran the remaining action when we only had 2, but now with our full specification it's much safer to have all 6 enumerated and error if we get something different. I can also add a test case to make sure we're covered in that eventuality, as well.

packages/index-lambda/tests/test_index_lambda_function.py
bamader: For sure!

packages/index-lambda/src/index_lambda/lambda_function.py
bamader: Mmm yeah great point. I think testing if `"result_cache" in action` is my preferred option here. Quick and easy to do, and also establishes a convention that if we add future actions, we should always be using `result_cache` explicitly in them, which I like!

packages/index-lambda/src/index_lambda/lambda_function.py
bamader: Good catch, I like that copy!

packages/index-lambda/src/index_lambda/lambda_function.py
bamader: Sure thing! I didn't even notice that was a pattern or something to consider, so good little learning note for me.

packages/index-lambda/tests/test_index_lambda_function.py
bamader: Oh lol you're completely right, I forget the additional `key` typing in the pair literally every time I do type declarations for dictionaries 🤦

packages/index-lambda/tests/test_index_lambda_function.py
bamader: Ah yep, I see what you mean. Can def change that text.

robertandremitchell (approved):
lgtm!

nickclyde (approved):
Thanks for the changes, looks good Brandon!!

BradySkylight (approved):
Nice!

Issue #547, opened 2026-05-12
Add versioned release workflow for Lambda Docker images on APHL ECR


nickclyde:
## Background

APHL's production change-control process requires semver-tagged container images plus written release notes for any promotion to production. The training pipeline can continue to track the floating `:latest` tag (current behavior), but production needs immutable `vX.Y.Z` tags it can pin to and a corresponding GitHub Release with notes.

Today neither of our existing workflows produces this:

- `.github/workflows/docker-image-push.yml` pushes `:latest` (and branch/SHA tags) to GHCR and APHL ECR, but no semver tag.
- `.github/workflows/release.yml` fires on `v*.*.*` git tag pushes and creates a GitHub Release, but its `package` job builds against a default `Dockerfile` at the repo root that doesn't exist, so no usable versioned image is produced, and nothing reaches APHL ECR. The version source script `dev_scripts/version_name.sh` is not coupled to anything authoritative.

## Goal

When a release author bumps `version` in the root `pyproject.toml` and merges to `main`, automation should:

1. Detect the version bump.
2. Create and push the matching `vX.Y.Z` git tag.
3. Build all three Lambda images (index, ttc, augmentation) for `linux/amd64`.
4. Push each image to APHL ECR with **both** `:vX.Y.Z` and `:latest` tags. (Also push to GHCR with the same tag set, matching the existing dev flow.)
5. Create a published GitHub Release named `vX.Y.Z` with auto-generated notes (`generate_release_notes: true` - lists merged PRs since the previous tag).
6. Post the release link to the existing Slack webhook (reuse the block in `release.yml`).

## Acceptance Criteria

- [ ] Single source of truth: root `pyproject.toml` `version` field drives the release tag.
- [ ] Bumping `version` in `pyproject.toml` on `main` (with no other changes required) results in a published GitHub Release and three `vX.Y.Z`-tagged images on APHL ECR within one workflow run.
- [ ] No release happens if the merged commit doesn't change `version`.
- [ ] No duplicate release if `vX.Y.Z` git tag already exists. The workflow should fail loudly rather than silently re-tag.
- [ ] `:latest` on APHL ECR is updated as part of the release so the training pipeline continues to receive the release build.
- [ ] Release notes on the GitHub Release list every PR merged since the previous `vX.Y.Z` tag.
- [ ] The existing dev push flow (push-to-main -> `:latest` + branch tag) remains intact for non-release commits, so engineers can keep iterating without cutting releases.
[...]

PR #549, opened 2026-05-12
Refresh Lambda base images and clear AWS Inspector findings

### LINKED ISSUE #548: Security: AWS Inspector findings on deployed Lambda images
## Summary

AWS Inspector v2 reports **31 high-severity and 21 medium/low findings** across the three currently deployed Lambda container images (`ttc-index-lambda`, `ttc-augmentation-lambda`, `ttc-lambda`). After deduplication these collapse to **19 unique CVEs** (11 High, 7 Medium, 1 Low). The same set is mostly common to all three images.

Every finding originates in the `public.ecr.aws/lambda/python:3.12` base image layers (OS packages, the bundled Python site-packages, and the `aws-lambda-rie` Go binary). None are caused by application code or workspace dependencies in `uv.lock`. The fix for ~17 of the 19 CVEs is to rebuild against a fresh base image; the rest need a small Dockerfile addition or a suppression with justification.

## Proposed fixes

1. **Rebuild against a fresh base image.** Add an explicit `docker pull public.ecr.aws/lambda/python:3.12` step (or `pull: true` on `docker/build-push-action`) before the build so the latest digest is fetched even with warm GHA cache. Optionally pin `FROM` to a specific digest and bump it on a recurring schedule. Expected to clear: all glibc, kernel-headers, `aws-lambda-rie`, `urllib3`, and `pip` findings (17 of 19 CVEs).

2. **Fix the lingering `setuptools` in `ttc-lambda`.** After the heavy `pip install torch …` step in `Dockerfile.ttc`, add `RUN pip install --no-cache-dir --upgrade setuptools>=78.1.1` so the final image carries a non-vulnerable version regardless of what an intermediate wheel pulled in. (Confirm afterwards that `uv.lock`'s pin of `setuptools 80.10.2` is not being shadowed by an earlier pip step.)

3. **Suppress CVE-2026-3219 with justification.** No upstream fix exists for this pip TAR/ZIP confusion finding. The image's `pip` is never invoked at Lambda runtime, so the practical risk is nil. Add an Inspector suppression rule (or document this in a security exceptions log) tied to this CVE with a 90-day review date.

4. **Add scheduled rebuilds.** Independent of code changes, schedule the `docker-image-push.yml` workflow on a weekly cron so newly published base-image patches land without waiting for the next feature deploy. This avoids a recurrence of the staleness pattern.

## Verification

After fixes 1–3 ship and the new images are deployed, recheck AWS Inspector findings.

Expected: `0` High remaining.

nickclyde:
## Summary
- `pull: true` + weekly cron on `docker-image-push.yml` so AWS-published `public.ecr.aws/lambda/python:3.12` patches actually land instead of being shadowed by warm GHA build cache
- Upgrade `setuptools` in `Dockerfile.ttc` after torch installs to clear CVE-2025-47273 (torch leaves an older copy that shadows the `uv.lock` pin)
- New `terraform/inspector.tf` suppresses CVE-2026-3219 (pip — no upstream fix, not on the Lambda runtime path) scoped to the three ECR repos, with a 90-day review date in the description

Closes #548.

## Notes

- The weekly cron pushes refreshed `latest`/`main` tags to GHCR/APHL ECR, but production Lambdas will be pinned to commit-SHA tags after #547, so let's make sure the new release workflow also includes `pull: true`.
- The Inspector suppression rule requires `terraform apply` after merge.

## Test plan
- [ ] Confirm CI build picks up a newer base image (compare base layer digest in build logs before/after)
- [ ] Once deployed, re-run AWS Inspector and verify 0 High findings remain on the three ECR images
- [ ] After `terraform apply`, confirm CVE-2026-3219 is suppressed (Medium count drops accordingly)
- [ ] Verify the Monday 10:00 UTC scheduled run fires

### DIFF
--- a/.github/workflows/docker-image-push.yml
+++ b/.github/workflows/docker-image-push.yml
@@ -4,6 +4,11 @@ on:
   push:
     branches: [main]
 
+  schedule:
+    # Weekly rebuild so AWS-published base-image patches land without
+    # waiting for a code change. Mondays 10:00 UTC (6 AM ET / 3 AM PT).
+    - cron: "0 10 * * 1"
+
   workflow_dispatch:
     inputs:
       ref:
@@ -79,6 +84,7 @@ jobs:
           context: .
           file: ${{ matrix.dockerfile }}
           push: true
+          pull: true
           platforms: linux/amd64
           provenance: false
           tags: ${{ steps.meta.outputs.tags }}
--- a/Dockerfile.ttc
+++ b/Dockerfile.ttc
@@ -12,6 +12,10 @@ RUN microdnf install -y gcc-c++ make && microdnf clean all
 RUN pip install --no-cache-dir torch==2.9.1 --index-url https://download.pytorch.org/whl/cpu
 RUN pip install --no-cache-dir huggingface_hub
 
+# Torch's build environment can leave behind an older setuptools that shadows
+# the workspace pin and trips CVE-2025-47273. Upgrade explicitly.
+RUN pip install --no-cache-dir --upgrade 'setuptools>=78.1.1'
+
 # Download retriever at build time (private repo, needs token)
 RUN --mount=type=secret,id=huggingface_token,env=HF_TOKEN \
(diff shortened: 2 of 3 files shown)


robertandremitchell (approved):
lgtm, 🤞🏽 it should clear the errors out

PR #553, opened 2026-05-13
refactor: simplify lambda handler functions by handling S3 client internally

JNygaard-Skylight:
## Description

Currently the lambda handler functions accept an S3 Client a parameter. However the Lambda Handler already creates a an S3 client internally and we do not have a use case where we would need to pass in a different client. Removing this parameter simplifies the Lambda handler, and the TTC and Augmentation Lambda functions.

## Related Issues

Closes # n/a

## Additional Notes

This comes out of the refactoring in [544](https://github.com/CDCgov/dibbs-text-to-code/pull/544).

### DIFF
--- a/packages/lambda-handler/src/lambda_handler/lambda_handler.py
+++ b/packages/lambda-handler/src/lambda_handler/lambda_handler.py
@@ -1,5 +1,5 @@
 import os
-import typing
+from typing import BinaryIO
 
 import boto3
 from aws_lambda_powertools import Logger
@@ -90,7 +90,7 @@ def create_s3_client() -> BaseClient:
     return _cached_s3_client
 
 
-def create_opensearch_client(aws_auth: AWS4Auth | None = None) -> OpenSearch:
+def create_opensearch_client() -> OpenSearch:
     """Creates an OpenSearch client.
 
     :param aws_auth: AWS4Auth object for authentication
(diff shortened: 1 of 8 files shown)


packages/lambda-handler/src/lambda_handler/lambda_handler.py
```diff
@@ -100,7 +100,7 @@ def create_opensearch_client(aws_auth: AWS4Auth | None = None) -> OpenSearch:
 
     if _cached_opensearch_client is None:
         endpoint_url = get_env_variable("OPENSEARCH_ENDPOINT_URL")
-        auth = aws_auth or create_aws_auth()
```
robertandremitchell, line 103: do we still need this for caching purposes? that was the original thing we were trying to check for with this logic

robertandremitchell: same with the other 2 vars

robertandremitchell (commented):
small question but otherwise makes sense to me

packages/lambda-handler/src/lambda_handler/lambda_handler.py
JNygaard-Skylight: `create_s3_client()`, `create_opensearch_client()`, `create_aws_auth()` all already have logic for checking if the client already exists.

robertandremitchell (approved):

PR #555, opened 2026-05-13
refactor: make all models frozen

JNygaard-Skylight:
## Description

Creates a base model that all other models inherit so that all models can be frozen and do not allow extra attributes. This matches how we are using the models. In addition it makes serialization/deserialization predictable as stray attributes cannot be added to models.

## Related Issues

Closes # n/a

## Additional Notes

This comes out of the refactoring in [544](https://github.com/CDCgov/dibbs-text-to-code/pull/544). By ensuring models are exactly what we expect we can simplify the serialization/deserialization.

### DIFF
--- a/packages/augmentation/src/augmentation/models/application.py
+++ b/packages/augmentation/src/augmentation/models/application.py
@@ -1,8 +1,6 @@
 from enum import Enum
 
-from pydantic import BaseModel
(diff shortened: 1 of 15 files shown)


robertandremitchell (approved):
makes sense to me

PR #557, opened 2026-05-14
refactor: Simplify get_relevant_text by getting the criteria itself

JNygaard-Skylight:
## Description

Currently the function `select_relevant_text` in `evaluator.py` requires that the criteria are passed into it as a parameter. The criteria are gotten from the `get_evaluation_criteria_for_data_field` in the same module and only requires that a data field type is passed in. There is no reason why the criteria needs to be a public function, and instead the data field type can just be passed into `select_relevant_text`. This simplifies the use `select_relevant_text` as only function needs to be called, and reduces the number of public functions in `evaluator.py`.

Removed tests that were directly testing private functions, and replaced them with tests that cover the same cases but using the public method `select_relevant_text`.

## Related Issues

Closes # n/a 
## Additional Notes

This comes out of the refactoring in [544](https://github.com/CDCgov/dibbs-text-to-code/pull/544).


packages/text-to-code/src/text_to_code/services/evaluator.py
```diff
@@ -107,6 +98,7 @@ def select_relevant_text(
     :param criteria: The evaluation criteria defining priority order and translation behavior.
```
nickclyde, line 98: I think this docstring is now stale, since the `criteria` parameter has been replaced with `field_type`. Could we update this to describe `field_type: The data field whose evaluation criteria should be used.` (or similar) so it matches the new signature?

packages/text-to-code/tests/unit/test_evaluator.py
```diff
@@ -431,28 +147,67 @@ def test_returns_none_when_all_candidates_are_blank_or_missing_for_priorities()
...
+    actual = select_relevant_text(candidates, DataField.LAB_TEST_NAME_RESULTED)
 
-def test_get_evaluation_criteria_for_data_field_raises_for_unregistered_data_field(mocker) -> None:
-    class EmptyRegistry(dict):
-        def __getitem__(self, key) -> KeyError:
-            raise KeyError(key)
+    assert actual == candidates[0]
+
+
+def test_select_relevant_text_no_systems(mocker):
```
nickclyde: It looks like `mocker` is in the signature but never used in this test. Same applies to `test_select_relevant_text_not_loinc_or_snomed_system` below. Could we drop the fixture from both since they no longer need to patch anything?

packages/text-to-code/tests/unit/test_evaluator.py
```diff
@@ -431,28 +147,67 @@ def test_returns_none_when_all_candidates_are_blank_or_missing_for_priorities()
...
-
-    selected = select_relevant_text(candidates=candidates, criteria=criteria)
+    selected = select_relevant_text(candidates, DataField.LAB_TEST_NAME_RESULTED)
 
     assert selected is None
 
 
-def test_get_evaluation_criteria_for_data_field_returns_criteria_instance() -> None:
-    criteria = get_evaluation_criteria_for_data_field(DataField.LAB_TEST_NAME_RESULTED)
+def test_select_relevant_text_selection_strategy_first(mocker):
```
nickclyde, line 155: Small thought: since there's only one candidate here and it has `system=None`, both `FIRST` and the default `PREFER_LOINC_THEN_SNOMED` strategies will return the same result via the `any_has_system` short-circuit in `_select_translation_candidate`. So I don't think this test actually exercises the FIRST branch in a way that would catch a regression. Would adding a second candidate with a recognized system (e.g. a LOINC one) make this test meaningfully distinguish FIRST from the default? Then we'd see FIRST return candidate[0] while PREFER_LOINC_THEN_SNOMED would return the LOINC one.

nickclyde (approved):
Nice cleanup! Reads much more naturally now. The collapsing of the criteria lookup into `select_relevant_text` is a good simplification and the test rewrites to drive everything through the public function are an improvement.

A few small things to consider:

PR #559, opened 2026-05-15
Adding code so eicr always flows to augmentation

### LINKED ISSUE #545: Add failsafe so original eICR always flows through to AugmentationEICRV2/
## Background

The APHL AIMS eCR pipeline assumes that any eICR diverted into the TTC pipeline will reappear at `s3://<bucket>/AugmentationEICRV2/<persistence_id>` so it can be picked up by downstream processing. Today this assumption is violated in several cases (when there are no relevant Schematron errors, when an exception is raised in TTC, or when an exception is raised in Augmentation) and the eICR is silently dropped from the main pipeline. APHL has asked for a failsafe that guarantees the original document always makes it through.

## Failure modes today

| Scenario | `AugmentationEICRV2/` populated? |
|---|---|
| Happy path (errors found, matches found) | ✅ Augmented eICR |
| No relevant Schematron errors | ❌ Nothing. TTC early-exits at `packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py:472-485` without writing a TTC output, so the Augmentation Lambda is never triggered |
| Unhandled exception in TTC Lambda | ❌ Nothing. caught at handler boundary (`lambda_function.py:69-75`), no output written |
| Unhandled exception in Augmentation Lambda | ❌ Nothing. caught at handler boundary (`packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py:52-58`), no output written |
| TTC found zero code matches | ⚠️ Augmentation Lambda runs and writes a file, but `EICRAugmenter.augment()` still rewrites document headers even with an empty codes list (`packages/augmentation/src/augmentation/services/eicr_augmenter.py:50-83`). The original is modified, not passed through unchanged. |

## Proposed behavior

The Augmentation Lambda becomes the single owner of writes to `AugmentationEICRV2/`. The TTC Lambda is updated to always emit a `TTCAugmentationMetadataV2/{persistence_id}` record, optionally carrying a `passthrough: true` flag and a structured `passthrough_reason`. The Augmentation Lambda, when it sees the flag (or hits its own exception), skips augmentation and writes the original eICR byte-for-byte to `AugmentationEICRV2/{persistence_id}`. An `AugmentationMetadataV2/{persistence_id}` record captures the reason for observability.

Passthrough reasons (enum):
- `no_relevant_schematron_errors`
- `no_code_matches`
- `ttc_exception`
- `augmentation_exception`

### TTC Lambda changes (`packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py`)

1. **No-relevant-errors branch (lines 472-485):** instead of writing only metadata and returning, also write a TTC output to `TTCAugmentationMetadataV2/{persistence_id}` with `passthrough: true` and `passthrough_reason: "no_relevant_schematron_errors"`. Keep `schematron_errors` empty. This triggers the Augmentation Lambda, which sees the flag and writes the original eICR.
2. **Handler-level exception (lines 69-75):** before adding to `failures`, attempt a best-effort write of a passthrough TTC output (`passthrough: true`, `passthrough_reason: "ttc_exception"`, optional `error` string for observability - string only, never a stack trace with PHI). Wrap in its own try/except so a secondary failure still logs cleanly. If `persistence_id` cannot be extracted from the S3 event, skip the passthrough write and let the SQS DLQ handle it.
3. **TTC-found-zero-matches path:** when the output has an empty `schematron_errors` map after processing, set `passthrough: true` and `passthrough_reason: "no_code_matches"` on the TTC output. Same downstream behavior.

### Augmentation Lambda changes (`packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py`)

1. **Detect passthrough in `_process_record` (around line 113):** after loading `ttc_output`, check `ttc_output.get("passthrough")`. If true, skip the augmenter entirely. Write the **original** eICR bytes unmodified to `AugmentationEICRV2/{persistence_id}` and an augmentation metadata file recording `{passthrough: true, passthrough_reason: <upstream reason>}`.
2. **Wrap augmentation in try/except (lines 130-138):** if `EICRAugmenter` raises, catch, log, and fall back to writing the original eICR unchanged with metadata `{passthrough: true, passthrough_reason: "augmentation_exception", error: <str>}`. Mark the SQS record as a success at the handler level, as the passthrough is the intended outcome.
3. **Pre-augmenter failures (S3 read of original eICR fails, TTC output missing, JSON parse fails):** leave alone. The existing handler-level catch + SQS DLQ (`maxReceiveCount: 3` in `terraform/main.tf`) handle these; if we can't read the original, there is nothing to pass through.

### Shared model changes (`packages/shared-models/`)

- Add optional `passthrough: bool` and `passthrough_reason: str | None` fields to the TTC output schema. Both Lambdas read via `.get(...)`, so the change is backward-compatible.
- Use an enum for the four reason values listed above.

## Acceptance criteria

- Any `TextToCodeSubmissionV2/<persistence_id>` that triggers the TTC Lambda results in a corresponding object at `AugmentationEICRV2/<persistence_id>` (augmented or original, byte-for-byte).
- Passthrough writes produce a structured `AugmentationMetadataV2/<persistence_id>` record with `passthrough: true` and `passthrough_reason`.
- Augmented-path behavior is unchanged.
- Structured log fields include `passthrough_reason` so we can build a failsafe-rate metric in CloudWatch.
- The only conditions under which `AugmentationEICRV2/<persistence_id>` is *not* populated are (a) the triggering S3 event has no recoverable `persistence_id`, or (b) the original eICR object is unreadable; both fall through to the existing SQS DLQ.

## Files in scope

- `packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py`
- `packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py`
- `packages/shared-models/`
- Tests in `packages/text-to-code-lambda/tests/` and `packages/augmentation-lambda/tests/`

No Terraform changes are required.

## Test plan

- **Unit tests** (`uv run pytest packages/text-to-code-lambda/tests/ packages/augmentation-lambda/tests/`): one test per passthrough reason. For passthrough cases, assert that bytes written to `AugmentationEICRV2/` are byte-for-byte equal to bytes read from `TextToCodeSubmissionV2/` (compare on the mocked S3).
- **End-to-end (local Docker Compose):** drop a clean eICR (no Schematron errors) into the input prefix, confirm a passthrough copy lands in `AugmentationEICRV2/`. Repeat with a deliberately malformed eICR to exercise the TTC exception path.
- **Lint/type:** `just ruff && just ty`.

robertandremitchell:
## Description

This adds a new passthrough logic to ttc-lambda to ensure that all eICR flow to AugmentationEICRV2/. This is necessary for APHL to pick up the eICR regardless of it needed augmentation and some uncovered exceptions.

ended up falling a little bit back into the e2e rabbit hole, as this change does impact (positively!) some "brittleness" I noticed/encountered on the last e2e ticket where augmented eicrs were getting tripped up because they didn't have any data needing augmenting. Now that we expect more eicrs to fully run end to end, I changed some of the logic there to reflect that. I can make that set of work a sub-branch of this ticket, if we prefer to review those two pieces of work separately.

## Related Issues

Closes #545 

## Additional Notes

## Checklist

## Checklist for Reviewers


packages/augmentation/src/augmentation/models/application.py
```diff
@@ -46,3 +47,5 @@ class TTCAugmenterOutput(FrozenBaseModel):
     persistence_id: str
     augmented_eicr: str
     metadata: Metadata
+    passthrough: bool = False
+    passthrough_reason: PassthroughReason | None = None
```
nickclyde: I think it might be worth pulling `passthrough` and `passthrough_reason` (and arguably `error`) onto `Metadata` instead of `TTCAugmenterOutput`. Right now `_save_augmentation_outputs` calls `output.metadata.model_dump()` and then manually adds the passthrough fields into the dict at write time. If those fields lived on `Metadata`, the serialization could stay a clean `model_dump_json()` and downstream consumers could re-validate without surprise. What do you think?

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -80,6 +82,26 @@ def handler(event: SQSEvent, context: LambdaContext) -> dict:
...
+    if passthrough_reason is None:
+        return None
+
+    if isinstance(passthrough_reason, PassthroughReason):
+        return passthrough_reason
+
+    if isinstance(passthrough_reason, str):
+        return PassthroughReason(passthrough_reason)
+
+    raise TypeError("passthrough_reason must be a string or PassthroughReason")
```
nickclyde: Small nit: this `TypeError` will bubble up to the outer `try` in `_process_record` and end up classified as `AUGMENTATION_EXCEPTION`, which would be a bit misleading since the underlying issue is malformed TTC output (not an augmentation failure). Could we either handle the unknown type inline (return `None` plus a `logger.warning`) so we still write a sensible passthrough?

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -107,36 +129,115 @@ def _process_record(record: SQSRecord) -> None:
...
+            # Currently only supports eICR augmentation. Other document types (e.g. from
+            # ecr-refiner or other services) may need different augmentation strategies.
+            config = TTCAugmenterConfig()
+            augmenter = EICRAugmenter(
+                document=original_eicr,
+                nonstandard_codes=augmenter_input.nonstandard_codes,
+                config=config,
+                deterministic_id_seed=augmenter_input.persistence_id,
+            )
+            original_eicr_id = str(augmenter.original_eicr_id)
```
nickclyde, line 187: I noticed we construct `EICRAugmenter` (which parses the XML) before checking the passthrough flag. That's intentional because we want `augmenter.original_eicr_id` in the metadata, but it means we pay parsing cost on every TTC-passthrough record, and a parser failure on a TTC-passthrough record gets reclassified as `AUGMENTATION_EXCEPTION`. Could we move the construction after the `if ttc_output.get("passthrough")` branch, and for the passthrough case extract the eICR ID separately (or just use `persistence_id` like the exception branch does)?

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -216,10 +325,18 @@ def _save_augmentation_outputs(
...
         status="success",
+        passthrough=output.passthrough,
+        passthrough_reason=output.passthrough_reason,
     )
 
+    metadata_output = output.metadata.model_dump()
+
+    if output.passthrough:
+        metadata_output["passthrough"] = output.passthrough
+        metadata_output["passthrough_reason"] = output.passthrough_reason
```
nickclyde: I think we should always include `passthrough` (and `passthrough_reason`) in the serialized metadata, not just when `output.passthrough` is `True`. Right now downstream consumers have to do `metadata.get("passthrough")` and treat missing as `False`, which is easy to miss. Always emitting `passthrough: false, passthrough_reason: null` for the success case would make the schema uniform. (This ties in with the model-level suggestion on `application.py:51` - if those fields lived on `Metadata`, `model_dump_json()` would do this automatically.)

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -97,6 +102,79 @@ def handler(event: SQSEvent, context: LambdaContext) -> dict:
...
+def _set_passthrough(
+    ttc_output: dict,
+    ttc_metadata_output: dict,
+    passthrough_reason: PassthroughReason,
+    error: str | None = None,
+) -> None:
+    ttc_output["passthrough"] = True
+    ttc_output["passthrough_reason"] = passthrough_reason.value
+    ttc_metadata_output["passthrough"] = True
+    ttc_metadata_output["passthrough_reason"] = passthrough_reason.value
```
nickclyde: Tiny consistency nit: here we write `passthrough_reason.value` (the raw string), but the augmentation lambda's `_save_augmentation_outputs` writes the `PassthroughReason` instance directly and relies on `json.dumps(default=str)` to coerce it. Both end up as the same JSON string (because `StrEnum` is a `str`), but it's a little easier to reason about if both lambdas pick one form. I'd suggest dropping `.value` and writing the enum directly in both places.

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -97,6 +102,79 @@ def handler(event: SQSEvent, context: LambdaContext) -> dict:
...
+
+        return True
+    except Exception:
+        logger.exception(
+            "Failed to write TTC exception passthrough output",
+            message_id=record.message_id,
+            status="error",
+            passthrough_reason=PassthroughReason.TTC_EXCEPTION.value,
+        )
+        return False
```
nickclyde, line 181: Should this function get a docstring?

e2e/test_e2e.py
```diff
@@ -330,54 +348,51 @@ def test_upload_and_process(
...
-        self._run_eicr_pipeline(
-            aws,
-            infra,
-            schematron_path,
-            eicr_path,
-            mock_lambda_context,
-        )
+            if passthrough_reason != PassthroughReason.AUGMENTATION_EXCEPTION:
+                assert ttc_output["passthrough"] is True
+                assert ttc_output["passthrough_reason"] == passthrough_reason
```
nickclyde, line 375: Question for my own understanding: why does the `AUGMENTATION_EXCEPTION` case skip the `ttc_output` check? Is it that the TTC lambda might have succeeded (with code matches) and only augmentation tripped, so TTC output won't be flagged as passthrough?

nickclyde (commented):
Nice work pulling the passthrough behavior together!

A few thoughts inline below, and also:

`e2e/test_e2e.py` - `_assert_s3_object_not_found` looks like dead code now, nothing in the file calls it now. Could we delete it?

packages/augmentation/src/augmentation/models/application.py
robertandremitchell: that makes sense to me. I can make this update.

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
robertandremitchell: yeah, I was having trouble thinking of how to write this part to preserve with the original_eicr_id. I can keep plugging on this. that method makes sense to me.

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
robertandremitchell: sounds good. this was a weird github copilot thing where it kept appending `.value` that I think I gave up and let the machines win

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
robertandremitchell: whoops. I'm kinda surprised it didn't throw a error

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
robertandremitchell: I think this is addressed, but curious your thoughts!

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
robertandremitchell: added

e2e/test_e2e.py
robertandremitchell: `AUGMENTATION_EXCEPTION` skips the TTC passthrough check because that check happens before augmentation is attempted. If TTC lambda requested passthrough, the function returns early. This branch only handles failures that occur after Augmentation begins processing a non-passthrough TTC output.

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
nickclyde: Looks great, thanks for addressing!

nickclyde (approved):
Thanks for tackling this Rob!!

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -80,6 +82,39 @@ def handler(event: SQSEvent, context: LambdaContext) -> dict:
...
+    """
+    passthrough_reason = ttc_output.get("passthrough_reason")
+
+    if passthrough_reason is None:
+        return None
+
+    if isinstance(passthrough_reason, PassthroughReason):
+        return passthrough_reason
+
+    if isinstance(passthrough_reason, str):
```
BradySkylight, line 99: This is cool!  I do wonder, will there be a request to this function that won't use the actual PassthroughReason and a string instead?

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -107,36 +142,90 @@ def _process_record(record: SQSRecord) -> None:
 
         ttc_output = _load_ttc_output(persistence_id, bucket_name)
         original_eicr = _load_original_eicr(persistence_id, bucket_name)
+
+        if ttc_output.get("passthrough"):
```
BradySkylight, line 146: This is just here to ensure there 'is' a passthrough reason before trying to do something with?  Maybe we use the function above and just ensure it's not None?  Meaning, just call _get_passthrough_reason(ttc_output) and add it to the metadata, if there isn't one then the passthrough_reason will be None.  Would that be ok?

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -185,9 +274,17 @@ def _parse_nonstandard_codes(ttc_output: dict) -> list[NonstandardCodeInstance]:
     :return: A list of NonstandardCodeInstance objects.
     """
     codes = []
-    for entries in ttc_output.get("schematron_errors", {}).values():
+    schematron_errors = ttc_output.get("schematron_errors", {})
```
BradySkylight, line 277: If there are no schematron_errors found then what is the passthroughreason?  Is this set somewhere else?

packages/shared-models/src/shared_models/__init__.py
```diff
@@ -63,3 +63,12 @@ class TTCAugmenterInput(FrozenBaseModel):
...
     nonstandard_codes: list[NonstandardCodeInstance]
+
+
+class PassthroughReason(StrEnum):
+    """Reasons why augmentation was bypassed and the original eICR was passed through."""
+
+    NO_RELEVANT_SCHEMATRON_ERRORS = "no_relevant_schematron_errors"
+    NO_CODE_MATCHES = "no_code_matches"
+    TTC_EXCEPTION = "ttc_exception"
+    AUGMENTATION_EXCEPTION = "augmentation_exception"
```
BradySkylight, line 74: Nice!

BradySkylight (commented):
A few minor questions and comments, but this looks good overall!

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
robertandremitchell: no, I think this is a leftover from when I was using `.value` in some parts to turn passthroughreason into a string.

robertandremitchell: I guess it's a nice guard still if somehow an invalid passthrough reason is passed through

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
BradySkylight: Found it further in the code.  Nice!

robertandremitchell: yeah, this is set further upstream in the ttc-lambda. if there are no schematron errors it gets set to `NO_RELEVANT_SCHEMATRON_ERRORS`

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
robertandremitchell: the main check is to determine whether a reason exists and if the passthrough reason means we ought to skip augmentation. passthrough reason is in the metadata, but not sure i follow what the edit would be. happy to discuss/change

BradySkylight: Got it.  That makes more sense.  I was missing the other areas in the code where the reason is being set.  Now that I see it, it makes sense all together.  No change needed.

BradySkylight (approved):
Looks good.  Nice job pulling this together so quick!

PR #562, opened 2026-05-15
Create result cache service

bamader:
## Description

This PR creates the result cache service within text-to-code and builds out its GET and PUT functionality for storing and retrieving cached results within open search. I opted to make this its own PR before integrating with e2e since there's a new service being created, and this helps keep each individual PR review small.

## Related Issues

Related to a task in #533

### DIFF
--- a/packages/text-to-code/src/text_to_code/services/result_cache.py
+++ b/packages/text-to-code/src/text_to_code/services/result_cache.py
@@ -114,7 +114,7 @@ def put_new_cached_result(
         cache_key=cache_key,
         text=candidate_input,
         data_field=data_field,
-        loinc_code=json.dumps(loinc_code),
+        loinc_code=json.dumps(loinc_code.__dict__),
         search_score=search_score,
         reranker_score=reranker_score,
         cached_at=datetime.now(UTC).isoformat(),
--- a/packages/text-to-code/src/text_to_code/services/result_cache.py
+++ b/packages/text-to-code/src/text_to_code/services/result_cache.py
@@ -76,9 +76,9 @@ def get_cached_result(
     :param index: The namespace of the Index to check.
     :param os_doc_id: The OpenSearch document `_id` parameter to check for.
     """
-    response: OpenSearchGetResponse = opensearch_client.get(index=index, id=os_doc_id)
+    response = opensearch_client.get(index=index, id=os_doc_id)
     if response and response["found"]:
-        return response.source
+        return response["source"]
     return None
 
 
--- a/packages/text-to-code/src/text_to_code/services/result_cache.py
+++ b/packages/text-to-code/src/text_to_code/services/result_cache.py
@@ -77,7 +77,7 @@ def get_cached_result(
     :param os_doc_id: The OpenSearch document `_id` parameter to check for.
     """
     response: OpenSearchGetResponse = opensearch_client.get(index=index, id=os_doc_id)
-    if response and response.found:
+    if response and response["found"]:
         return response.source
     return None
 
--- a/packages/text-to-code/src/text_to_code/services/result_cache.py
+++ b/packages/text-to-code/src/text_to_code/services/result_cache.py
@@ -1,0 +1,121 @@
+import json
+from datetime import UTC
+from datetime import datetime
+from hashlib import sha256
+
+from opensearchpy import OpenSearch
+from pydantic import Field
+
+from shared_models import Code
+from shared_models import FrozenBaseModel
+
+
+class OpenSearchResultCacheSource(FrozenBaseModel):
+    """Represents a term-standardization mapping returned from the OpenSearch Result Cache Index."""
+
+    cache_key: str = Field(
+        description="The computed hash key mapping candidate text + data field to this known hit."
+    )
+    text: str = Field(description="The input candidate text this Cache Result is associated with.")
+    data_field: str = Field(
+        description="The data field from which the candidate text was retrieved during initial "
+        "processing of this cache hit."
+    )
+    loinc_code: str = Field(description="The serialized Code object for this cached input.")
+    search_score: float = Field(
+        description="The cosine similarity score calculated for this standardization during "
+        "initial processing of this input."
+    )
+    reranker_score: float = Field(
+        description="The cross-encoder score calculated for this standardization during initial "
+        "processing of this input."
+    )
+    cached_at: str = Field(description="The timestamp at which this input was cached.")
+
+
+class OpenSearchGetResponse(FrozenBaseModel):
+    """Represents the response body returned by an ID-indexed GET request to the OpenSearch API."""
+
+    index: str = Field(description="The name of the index containing the document.", alias="_index")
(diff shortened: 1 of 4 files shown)


packages/text-to-code/src/text_to_code/services/result_cache.py
```diff
@@ -0,0 +1,129 @@
...
+    )
+    routing: str = Field(
+        description="The routing value used to determine which shard stores the document. "
+        "Only included if a routing value was specified when the document was indexed.",
+        alias="_routing",
+    )
+    source: OpenSearchResultCacheSource = Field(description="The cached result for the input.")
+    fields: dict[str, list] = Field(
+        description="Stored field values of the mappings.", alias="_fields"
+    )
```
robertandremitchell: based on structure of the other files (whole other story maybe 😅 ), I think these should live in a separate file in the `models/` directory

robertandremitchell (commented):
makes sense to me. One small note about moving the classes.

packages/text-to-code/src/text_to_code/services/result_cache.py
```diff
@@ -0,0 +1,129 @@
...
+        (candidate_input.strip().lower() + "|" + data_field).encode("utf-8")
+    ).hexdigest()
+    new_cache_hit: OpenSearchResultCacheSource = OpenSearchResultCacheSource(
+        cache_key=cache_key,
+        text=candidate_input,
+        data_field=data_field,
+        loinc_code=json.dumps(loinc_code.__dict__),
+        search_score=search_score,
+        reranker_score=reranker_score,
+        cached_at=datetime.now(UTC).isoformat(),
```
BradySkylight: Do we have an idea of how long to keep these results?  Not sure we need to 'solve' this for this PR.

packages/text-to-code/src/text_to_code/services/result_cache.py
```diff
@@ -0,0 +1,129 @@
...
+        alias="_seq_no",
+    )
+    primary_term: int = Field(
+        description="The primary term assigned to the document for the indexing operation. "
+        "Used with `seq_no` for optimistic concurrency control.",
+        alias="_primary_term",
+    )
+    found: bool = Field(
+        description="Indicates whether the document exists (`True` if it was found)."
+    )
```
BradySkylight: This is specifically if there was a match found in Opensearch for the text/vector incoming in question, correct?

packages/text-to-code/src/text_to_code/services/result_cache.py
```diff
@@ -0,0 +1,129 @@
...
+    :param os_doc_id: The OpenSearch document `_id` parameter to check for.
+    """
+    response = opensearch_client.get(index=index, id=os_doc_id)
+    if response and response["found"]:
+        return response["source"]
+    return None
+
+
+def put_new_cached_result(
+    opensearch_client: OpenSearch,
```
JNygaard-Skylight, line 30: This is a lot of parameters, could this be converted to a single object? I suppose regardless we still need to bring all of these into a single thing, so it probably just moves the problem around. Idk, something to think about.

packages/text-to-code/src/text_to_code/services/result_cache.py
```diff
@@ -0,0 +1,129 @@
...
+    loinc_code: str = Field(description="The serialized Code object for this cached input.")
+    search_score: float = Field(
+        description="The cosine similarity score calculated for this standardization during "
+        "initial processing of this input."
+    )
+    reranker_score: float = Field(
+        description="The cross-encoder score calculated for this standardization during initial "
+        "processing of this input."
+    )
+    cached_at: str = Field(description="The timestamp at which this input was cached.")
```
JNygaard-Skylight: This might just be a personal preference, but I think it would make sense to give this a default value of `datetime.now(UTC).isoformat()`. I don't think we would ever have a reason to set it to be anything else, so giving it a default removes a parameter to pass in, but it will still give us the option to if we need to.

packages/text-to-code/src/text_to_code/services/result_cache.py
bamader: Lol totally fair on both counts 😅 . I'll get those moved over!

packages/text-to-code/src/text_to_code/services/result_cache.py
bamader: I believe the idea is that we can keep these results until we perform any kind of update. If we update the model, the embeddings and search capabilities will change, so we'll need to be able to search fresh. And then if LOINC updates existing codes or releases new ones, we'll need the ability to possibly map existing results to them if they're better matches. The Terraform PR I just landed ties the bootstrapping of the result cache to our index lambda, so anytime we run the create there we'll build the cache fresh to keep things easy on ourselves.

packages/text-to-code/src/text_to_code/services/result_cache.py
bamader: Yeah, the "document" here just means whatever we're storing in OpenSearch. So when we use this on the result cache, this will be a cache-key hit with an object holding the known standard LOINC. This class structure also defines the general API response so we could absolutely use this for other indexing too. I'll update all this in the documentation to make it clearer.

packages/text-to-code/src/text_to_code/services/result_cache.py
bamader: Yeah I actually had an object at first, but I really didn't like the way it turned out. If we were to make an object that the main lambda invoked on each query, we take a small performance hit for object creation, still have to use all these parameters (so you're right it just gets pushed elsewhere), and then we also wind up with a pretty  bloated class that doesn't get re-used elsewhere. I think the cleanest and most efficient method is unfortunately to just pass in the parameters and accept that _someone_ has to be responsible for the aggregation, so it might as well be the only function using this particular structure of information.

packages/text-to-code/src/text_to_code/services/result_cache.py
bamader: Sure, can def do that

pyproject.toml
```diff
@@ -95,10 +95,11 @@ select = [
   "W",    # pycodestyle warning
 ]
 ignore = [
-  "D100",   # undocumented-public-module
-  "E501",   # line-too-long - Handled by formatter
-  "RET504", # unnecessary-assign
-  "S314",   # recommend `defusedxml` over `xml` and `lxml`
+  "D100",    # undocumented-public-module
+  "E501",    # line-too-long - Handled by formatter
+  "PLR0913", # too many arguments in function definition
```
JNygaard-Skylight: While I agree with the rationale for using several arguments for `put_new_cached_result`, I still think this is a good linting rule and should not be disabled globally. Perhaps we can just disable it inline with a small comment with our rationale?

bamader: I can switch to inline disable. I personally dislike this rule but not a hill I'll die on, so inline is fine.

BradySkylight (approved):
With the answers to my questions, that help me clarify my own understanding, This makes sense and looks good to me.

PR #568, opened 2026-05-20
Extending schematron validation in e2e tests

robertandremitchell:
## Description

Enhances e2e tests to meet acceptance criteria from ticket:

* E2E tests verify augmented eICR XML is well-formed
* E2E tests verify augmented eICR contains expected <translation> elements
* E2E tests verify augmented eICR retains all original content (no data loss)
* Stretch: validate against actual Schematron rules if available

The bulk of the changed files are about bring validation logic into the augmentation-lambda. Is this something we want to do? I fell down this rabbit hole with the e2e tests because one of them was invalid. Based on the ticket, it sounds like APHL wants Mirth to do this? In which case, maybe we don't need/want this code in the augmentation-lambda as well?

The alternative is just to drop this sample7 data from the e2e and replace it with another test where we expect to go e2e, or write a separate e2e pathway for it where we expect the data to get kicked back.

The work to complete integration of these rules into AWS will continue in #586.

## Related Issues

Closes #434 

## Additional Notes

## Checklist

## Checklist for Reviewers


packages/validation/src/validation/__init__.py
```diff
@@ -1,3 +1,3 @@
-from .main import validate_eicr
+from .main import ValidationResult, validate_eicr
```
nickclyde, line 1: I think there is a mismatch between the exported `ValidationResult` class and what `validate_eicr` actually returns. The class declares `error_id: str` and `locations: list[str]` (plural list), but `validate_eicr` builds and returns `dict`s shaped `{"error_id": ..., "location": ...}` (singular string). The test in `test_augmentation_lambda_function.py` ends up doing `cast(list[ValidationResult], [<dict>])` which makes the mismatch a bit visible.

packages/shared-models/src/shared_models/__init__.py
```diff
@@ -71,3 +71,4 @@ class PassthroughReason(StrEnum):
     NO_CODE_MATCHES = "no_code_matches"
     TTC_EXCEPTION = "ttc_exception"
     AUGMENTATION_EXCEPTION = "augmentation_exception"
+    AUGMENTATION_VALIDATION_FAILED = "augmentation_validation_failure"
```
nickclyde: Nit: the identifier is `AUGMENTATION_VALIDATION_FAILED` but the value is `"augmentation_validation_failure"`. Can we rename to `AUGMENTATION_VALIDATION_FAILURE`

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -217,6 +247,23 @@ def _process_record(record: SQSRecord) -> None:
             )
             return
 
+        validation_results = validate_eicr(output.augmented_eicr)
```
nickclyde: It seems that `validate_eicr` currently catches every exception internally and returns `[]` on failure (see `packages/validation/src/validation/main.py:98-99`). I think that means if Saxon ever fails to initialize, or the augmented XML can't be parsed, we will treat the result as "no errors" and ship a possibly broken eICR as success. Could we either tighten `validate_eicr` so callers can distinguish "validator threw" from "no errors", or wrap this call in a try/except that routes to `AUGMENTATION_VALIDATION_FAILED` (or `AUGMENTATION_EXCEPTION`) on unexpected failures?

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -111,6 +112,35 @@ def _get_passthrough_reason(ttc_output: dict[str, object]) -> PassthroughReason
...
+
+    :param persistence_id: The persistence ID for the eICR being processed.
+    :param original_eicr: The original eICR XML string.
+    :param attempted_output: The attempted augmentation output that failed validation.
+    :param validation_results: The validation errors returned by validate_eicr.
+    :return: The passthrough augmentation output.
+    """
+    metadata = Metadata(
+        original_eicr_id=attempted_output.metadata.original_eicr_id,
+        augmented_eicr_id=attempted_output.metadata.original_eicr_id,
```
nickclyde, line 135: Should `augmented_eicr_id` be the original ID here? When we are doing a passthrough we are writing the original eICR back out, so I think pointing both fields at `original_eicr_id` is correct, but it might be worth a short comment to make that explicit, since the `AUGMENTATION_EXCEPTION` path above does the same thing (`fallback_eicr_id = original_eicr_id or persistence_id`)

e2e/test_e2e.py
```diff
@@ -41,6 +41,14 @@
 BASE_FOLDER = Path(__file__).parent
 ASSETS_FOLDER = BASE_FOLDER / "assets"
 
+CDA_NAMESPACE = "urn:hl7-org:v3"
+CDA_NAMESPACES: dict[str, str] = {"cda": CDA_NAMESPACE}
+REGENERATED_DOCUMENT_HEADER_TAGS: tuple[str, str] = (
+    f"{{{CDA_NAMESPACE}}}id",
+    f"{{{CDA_NAMESPACE}}}effectiveTime",
+)
```
nickclyde, line 52: I'm not sure, but I think `setId` and `versionNumber` need to be in this tuple too. Looking at `packages/augmentation/src/augmentation/services/eicr_augmenter.py:100-122`, the augmenter regenerates four ClinicalDocument header children: `id`, `effectiveTime`, `setId`, and `versionNumber`. With only `id` and `effectiveTime` excluded, `_assert_augmented_eicr_retains_original_content` will treat the original `<setId>` / `<versionNumber>` signatures as missing whenever the augmenter actually changes them, and the test will fail on any fixture that exercises that code path. The fact that the suite is currently green is probably because most fixtures end up in the passthrough branch (where this assertion is skipped); we could end up tripping on the next fixture that goes end-to-end.

e2e/test_e2e.py
```diff
@@ -368,16 +516,24 @@ def test_upload_and_process(
...
                 PassthroughReason.TTC_EXCEPTION,
                 PassthroughReason.AUGMENTATION_EXCEPTION,
+                PassthroughReason.AUGMENTATION_VALIDATION_FAILED,
             ]
 
             if passthrough_reason != PassthroughReason.AUGMENTATION_EXCEPTION:
-                assert ttc_output["passthrough"] is True
-                assert ttc_output["passthrough_reason"] == passthrough_reason
+                assert augmentation_metadata.get("passthrough") is True
+                assert augmentation_metadata.get("passthrough_reason") == passthrough_reason
```
nickclyde: I think these two assertions are now no-ops. We are already inside `if augmentation_metadata.get("passthrough"):` (so `passthrough` is truthy), and `passthrough_reason` is defined two lines up as `augmentation_metadata["passthrough_reason"]`, so `augmentation_metadata.get("passthrough_reason") == passthrough_reason` is guaranteed. I think this was useful in the old code because we were comparing the TTC Lambda's output against the augmentation Lambda's output and confirming they agreed on the reason. Now that we only look at one dict, we could either delete this inner block, or keep the original cross-Lambda agreement check by also reading `ttc_output` again.

nickclyde (commented):
Nice work on the e2e coverage. The `_assert_augmented_eicr_retains_original_content` helper is really useful.

RE: Should this validation belong in the augmentation Lambda? - Two thoughts: (a) running schematron in-Lambda is useful even if Mirth/APHL also runs it, because it stops a malformed eICR from leaking out of our side and we get visibility (the `passthrough_reason` + `error` metadata) instead of a downstream 422 we have to chase. (b) Cold-start cost: `validate_eicr` invokes Saxon and compiles three XSLTs on the first call. The function caches outputs to disk, so warm invocations should be cheap. I think it's worth leaving in, but it would be good to confirm the cold-start delta on the deployed Lambda.

BradySkylight (commented):
Just at a higher level, are we able to verify that the original eICR, prior to Augmentation, has TTC errors and then after augementation, if expected, those TTC errors are now gone?  Maybe I am being too simplistic and/or you may actually be doing that (this is prior to me diving deeper into the code, but just read your PR overview)... I'm thinking like an error count check, a TTC error check before and after with expected count totals and TTC there before and not after (or sometimes if no code is found the TTC error would still be there.

packages/validation/src/validation/__init__.py
robertandremitchell: that makes sense to me. I think I got lost in the sauce that we'd have a list of ValidationResults but each one would just have an id and string location, regardless. updating!

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
robertandremitchell: updated doc block to make this clearer. Not sure if there should be another note

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py

e2e/test_e2e.py
```diff
@@ -147,6 +157,149 @@ def _drain_sqs_for_prefix(sqs_client, queue_url, prefix, max_messages=10) -> lis
...
+    """
+    parent = element.getparent()
+
+    if parent is None:
+        return False
+
+    if parent.tag != f"{{{CDA_NAMESPACE}}}ClinicalDocument":
+        return False
+
+    return str(element.tag) in REGENERATED_DOCUMENT_HEADER_TAGS
```
BradySkylight, line 200: This is cool!

e2e/test_e2e.py
```diff
@@ -147,6 +157,149 @@ def _drain_sqs_for_prefix(sqs_client, queue_url, prefix, max_messages=10) -> lis
...
+
+    for translation in translations:
+        assert etree.QName(translation).namespace == CDA_NAMESPACE, eicr_id
+
+
+def _assert_augmented_eicr_retains_regenerated_document_header_elements(
+    augmented_root: etree._Element,
+    eicr_id: str,
+) -> None:
+    """Assert that the augmented eICR retains the document header elements that the augmenter regenerates, even though their content may differ.
```
BradySkylight: Is this just ensuring the DocumentId, SetId, VersionNumber, and EffectiveDate are all still present, but doesn't care about the actual values?  This makes sure that the augmented eicr has these elements?

BradySkylight:
> Just at a higher level, are we able to verify that the original eICR, prior to Augmentation, has TTC errors and then after augementation, if expected, those TTC errors are now gone? Maybe I am being too simplistic and/or you may actually be doing that (this is prior to me diving deeper into the code, but just read your PR overview)... I'm thinking like an error count check, a TTC error check before and after with expected count totals and TTC there before and not after (or sometimes if no code is found the TTC error would still be there.

Just to add to this... I was assuming that the Schematron Validation was added into the E2E tests and that you were leveraging it here.  It seems that is not the case, but contains specific tests to ensure the augmentation is doing what is expected.  I think that makes sense and provides even more robust validation for Augmentation.

e2e/test_e2e.py
```diff
@@ -147,6 +157,149 @@ def _drain_sqs_for_prefix(sqs_client, queue_url, prefix, max_messages=10) -> lis
...
+    return signature_counts
+
+
+def _assert_augmented_eicr_contains_expected_translations(
+    augmented_root: etree._Element,
+    eicr_id: str,
+) -> None:
+    """Assert that the augmented eICR contains <translation> elements in the CDA namespace, which the augmenter is expected to inject.
+
+    :param augmented_root: The root element of the augmented eICR XML tree.
```
BradySkylight: Is this the root of the augmented eicr or is this the root where the augmentation occurred?  I think it should be the latter (ie..  the 'Observation' tag).  Am I thinking about this correctly?

e2e/test_e2e.py
robertandremitchell: updated!

robertandremitchell:
> Just at a higher level, are we able to verify that the original eICR, prior to Augmentation, has TTC errors and then after augementation, if expected, those TTC errors are now gone? Maybe I am being too simplistic and/or you may actually be doing that (this is prior to me diving deeper into the code, but just read your PR overview)... I'm thinking like an error count check, a TTC error check before and after with expected count totals and TTC there before and not after (or sometimes if no code is found the TTC error would still be there.

yeah, that's an interesting way to consider it. I think I was viewing this PR as independent i.e., reviewing an augmented eicr for any potential schematron errors, but it could be useful compare/use the errors that come from the initial round of validation and this second round.

e2e/test_e2e.py
robertandremitchell: augmented_root is the root of the full augmented eICR document

robertandremitchell: I can look at this aspect again

e2e/test_e2e.py
robertandremitchell: yeah, it's a check of presence only currently

e2e/test_e2e.py
BradySkylight: I think I may have just mis-understood between the docstring and the code.  Does the code below get 'all' translations no matter the 'descendants' in the path... meaning it just finds all 'translation' tags no matter the hierarchical depth? or is it only looking at the next level from the root for translation tags?  

`
translations: list[etree._Element] = augmented_root.xpath(
        "//cda:translation", namespaces=CDA_NAMESPACES
    )
`

robertandremitchell: I think the update I made should make the test more comprehensive in checking each translation

BradySkylight (approved):
Thanks for answering my questions and adding some changes.  This looks good!

nickclyde (approved):
Looks good, thanks Rob!!

PR #570, opened 2026-05-21
feat!: Refactor TTC lambda function and condense saved models.

### LINKED ISSUE #527: Missing data in TTC metadata
## Summary
The TTC metadata is saved as JSON in S3. This includes the string used to find a standardized code, the `candidate`, and results returned by the OpenSearch query, etc.

## Problem
1) The `candidate` is always saved an empty object: `"candidate": {},`
2) The `opensearch_processed_results` field is a string of the variable name: `"opensearch_retrieved_scores": "<opensearch_retrieved_scores>",`

### Example 
This is snapshot used by the `test_handler_success` test, `packages/text-to-code-lambda/tests/snapshots/test_lambda_function/test_handler_success/handler_success_ttc_metadata_output.json`:
```
{
  "eicr_metadata": "eicr_id=CdaInstanceIdentifier(null_flavor=None, assigning_authority_name=None, displayable=None, root='c8516bdc-8bb2-40aa-8dae-20a77546488f', extension=None) eicr_vendor='Test eCR Vendor Name'",
  "persistence_id": "<persistence_id>",
  "processed_at": "<processed_at>",
  "schematron_errors": {
    "Lab Test Name Ordered": [
      {
        "candidate": {},
        "eicr_id": null,
        "error": "Text to Code: Lab Test Name Ordered does not have a @code attribute",
        "error_context": "/ClinicalDocument/component[1]/structuredBody[1]/component[5]/section[1]/entry[1]/observation[1]",
        "error_id": null,
        "error_message": "Text to Code: Lab Test Name Ordered does not have a @code attribute",
        "error_test": " not(cda:code) or cda:code/@code or cda:code/cda:translation/@code",
        "field": "Lab Test Name Ordered",
        "opensearch_retrieved_scores": "<opensearch_retrieved_scores>",
// ...
```
## Expected Behaviour
1) The `candidate` field should be a `Candidate` object
2) The `opensearch_retrieved_scores` field should save the actual `OpenSearchResult` object

### Example
```

```

JNygaard-Skylight:
## Description
The primary goal of this branch was to remove fields from the models that are the outputs of our lambda functions (the metadata and the data that is passed from TTC to augmentation). This simplifies the code in several places and Pydantic handles the serialization/deserialization and allows us to improve the typing .

### Details
- Format snapshots to improve maintainability.
- Simplify the reading of the Augmentation input by loading it as a model, instead of just a dictionary. This removes the need for the `_parse_nonstandard_codes` function and all of its associated tests.
- Simplify the augmentation metadata:
  - Removed the `"schematron_error"` from the nonstandard code model
- Simplify the output of the TTC (the Augmentation input) by removing fields not used by the augmenter:
  -`"eicr_metadata"` field.
  - `"schematron_error"`
 - Combined two tests in `test_augmentation_lambda_function.py`: `test_handler_success` , and `test_handler_writes_outputs_to_s3` as they were testing the same thing, but with different asserts.
 - Created a new model for the TTC metadata, instead of a dictionary `TTCMetadata`.
 - Rewrote `_process_record_pipeline`:
   - Simplify the flow of data. Previously dictionaries were being implicitly passed between functions
   - Added type hinting where it was missing.

## Related Issues

Closes #527


packages/text-to-code-lambda/tests/conftest.py
```diff
@@ -112,11 +112,9 @@ def mock_aws_setup(monkeypatch: pytest.MonkeyPatch) -> boto3.client:
         s3.persistence_id = TEST_PERSISTENCE_ID
 
         # Put test Schematron error file in the mock S3 bucket
-        current_dir = Path(__file__).parent.parent.parent
-        schematron_path = (
-            current_dir / "text-to-code" / "tests" / "assets" / "test_schematron_errors.xml"
-        )
-        with schematron_path.open() as f:
+        with Path(
```
robertandremitchell: not entirely sure if this is os-agnostic. probably worth a separate cleanup tech debt ticket to check around some of these things

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -522,71 +391,141 @@ def _process_record_pipeline(
...
+
+    ttc_output = TTCAugmenterInput(
+        persistence_id=persistence_id,
+        nonstandard_codes=nonstandard_code_replacements,
+        passthrough=passthrough_reason is not None,
+        passthrough_reason=passthrough_reason,
+    )
+    ttc_metadata = TTCMetadata(
+        persistence_id=persistence_id,
+        eicr_metadata=processor.eicr_metadata,
```
nickclyde, line 448: I think this reordering introduces a behavior regression for malformed eICRs that have no relevant Schematron errors. Previously we returned early with `NO_RELEVANT_SCHEMATRON_ERRORS` before ever touching the eICR. Now we always construct `EicrProcessor` here (line 397) and access `processor.eicr_metadata` below (line 503), which raises `ValueError("No ID element found in eICR XML.")` for eICRs lacking a top-level `<id>` (e.g. `e2e/assets/eicr_empty/eicr_empty.xml`). That exception gets caught by the handler and rewritten as `TTC_EXCEPTION`, which is exactly what the eicr_empty snapshot diff shows (`no_relevant_schematron_errors` → `ttc_exception`).

This means eICRs that previously passed through cleanly as a no-op will now be marked as a system error.

Can we restore the early return when `schematron_data_fields` is empty, before touching the eICR?

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -161,14 +128,27 @@ def _process_record(record: SQSRecord) -> None:
...
+            output = TTCAugmenterOutput(
+                persistence_id=persistence_id,
+                augmented_eicr=original_eicr,
+                metadata=metadata,
+            )
+            _save_augmentation_outputs(persistence_id, output, bucket_name)
+            logger.info(
+                "Augmentation processing completed",
+                status="passthrough",
+                passthrough_reason=PassthroughReason.AUGMENTATION_EXCEPTION,
```
nickclyde: I think this log message might be incorrect. At this point the passthrough is happening because `nonstandard_codes` is empty, which is driven by whatever the upstream `passthrough_reason` was (NO_CODE_MATCHES, etc.), not AUGMENTATION_EXCEPTION. Could we log `passthrough_reason=passthrough_reason` (the local variable we just read on line 132) to match the other branches?

Also, given the upstream TTC lambda always sets `passthrough=True` whenever `nonstandard_codes` is empty, this branch should be unreachable in practice. Is it intentional as a defensive guard, or could we drop it and let the augment block handle the empty list?

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -30,10 +44,35 @@
...
+
+class TTCMetadata(FrozenBaseModel):
+    """Model to hold metadata about the TTC process."""
+
+    persistence_id: str
+    eicr_metadata: EICRMetadata | None = None
+    ttc_schematron_issues: list[TTCSchematronIssueDetail] | None = None
+    processed_at: datetime = Field(default_factory=lambda: datetime.now(UTC))
+    passthrough: bool | None = None
+    passthrough_reason: str | None = None
```
nickclyde: Tiny consistency nit: should this be `PassthroughReason | None = None`? The corresponding field on `TTCAugmenterInput` and `Metadata` is typed that way, and we're already passing `PassthroughReason` values into this field everywhere. Using `str | None` here loses the enum validation Pydantic would otherwise give us on deserialization.

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -522,71 +391,141 @@ def _process_record_pipeline(
...
+    if schematron_data_fields:
+        nonstandard_code_replacements: list[NonstandardCodeInstance] = []
+        ttc_schematron_issues_details: list[TTCSchematronIssueDetail] = []
+        for error in schematron_data_fields:
+            new_translation = None
+            unmatched_message = None
+            data_field = error.field
+            opensearch_retrieved_scores = None
+            ranked_results: list[ScoredResult] | None = None
+            passthrough_reason: PassthroughReason | None = None
```
nickclyde: I think this initialization shouldn't be in the for loop. The`passthrough_reason` is reset to `None` on every iteration of the loop but never mutated inside the loop body. It only gets a meaningful value after the loop (line 488 / 493). Could we move the initialization outside the `if schematron_data_fields:` block, before the loop??

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -1,24 +1,38 @@
...
 from text_to_code.services.embedder import embed
 from text_to_code.services.query import QueryBuilder
-from text_to_code.services.reranker import rerank
+from text_to_code.services.reranker import ScoredResult, rerank
 
 # Initialize the logger
 logger = Logger(service="ttc")
+# Forward standard logs into Powertools' structured output
+logging.getLogger("text_to_code").setLevel(logging.INFO)
+logging.getLogger("text_to_code").addHandler(logger.handlers[0])
```
nickclyde: `logger.handlers[0]` could raise `IndexError` if Powertools hasn't attached a handler yet (e.g. in some unit-test contexts where the logger is bypassed). Would it be simpler to keep the operational log lines ("Embedding the relevant text strings…", "Evaluating candidates…") in the lambda itself where the Powertools context is available, instead of inside the service modules?

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -314,133 +298,22 @@ def _build_nonstandard_code_instance(
...
-    ttc_metadata_output: dict,
+    metadata_output: TTCMetadata,
     bucket_name: str,
 ) -> None:
     """Save TTC metadata output to S3.
 
     :param persistence_id: The persistence ID extracted from the S3 object key
-    :param ttc_metadata_output: The TTC metadata output dictionary.
     :param bucket_name: The S3 bucket name to write to.
+    :metadata_output: The metadata model to be saved.
```
nickclyde: Typo: `:metadata_output:` should be `:param metadata_output:`

packages/text-to-code-lambda/tests/conftest.py
```diff
@@ -125,8 +123,7 @@ def mock_aws_setup(monkeypatch: pytest.MonkeyPatch) -> boto3.client:
         )
 
         # Put test eCR message file in the mock S3 bucket
-        ecr_path = current_dir / "text-to-code" / "tests" / "assets" / "basic_test_eicr.xml"
-        with ecr_path.open() as f:
+        with Path("packages/text-to-code-lambda/tests/assets/test_eicr.xml").open() as f:
```
nickclyde: These hardcoded paths assume the test runner's CWD is the repo root. The previous code used `Path(__file__).parent.parent.parent / "text-to-code" / "tests" / "assets" / ...`, which is CWD-independent. While `just test` works fine, it'd be safer to switch back to `Path(__file__).parent / "assets" / "test_schematron_errors.xml"` (and similarly for `test_eicr.xml`) so the tests can be invoked from any directory.

packages/shared-models/src/shared_models/__init__.py
```diff
@@ -57,17 +55,19 @@ class NonstandardCodeInstance(FrozenBaseModel):
...
     NO_CODE_MATCHES = "no_code_matches"
     TTC_EXCEPTION = "ttc_exception"
     AUGMENTATION_EXCEPTION = "augmentation_exception"
+
+
+class TTCAugmenterInput(FrozenBaseModel):
+    """Input for the augmentation service."""
+
+    persistence_id: str
+    nonstandard_codes: list[NonstandardCodeInstance] | None = None
```
nickclyde: Is the `| None` here intentional? In the TTC lambda it looks like we write either an empty list (when codes were searched but none matched) or a populated list. I don't see a code path that writes `None`. Making this `list[NonstandardCodeInstance] = Field(default_factory=list)` (or even keeping it as `list[NonstandardCodeInstance]` with a default of `[]`) would simplify: the augmentation lambda currently has `if not augmenter_input.nonstandard_codes:` which has to handle both `None` and `[]`. Defaulting to `[]` would let us use `len(...) == 0`.

packages/text-to-code-lambda/tests/test_lambda_function.py
```diff
@@ -1,134 +1,77 @@
...
+            },
+        ]
         mocker.patch(
-            "text_to_code.services.evaluator.select_relevant_text",
-            return_value=type("SelectedCandidate", (), selected_candidate)(),
+            "text_to_code_lambda.lambda_function.rerank",
+            return_value=ranked_results,
         )
 
+        """Test handler with no failures."""
```
nickclyde: Shouldn't this statement be the first line of the test?

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -30,10 +44,35 @@
 OPENSEARCH_ENDPOINT_URL = os.getenv("OPENSEARCH_ENDPOINT_URL")
 OPENSEARCH_INDEX = os.getenv("OPENSEARCH_INDEX", "ttc-index")
 
-# Constants
-NO_DATA_FIELDS_MESSAGE = (
-    "No relevant data fields identified from Schematron errors for TTC processing"
-)
+
+class TTCSchematronIssueDetail(FrozenBaseModel):
```
robertandremitchell: should this live in the `models` part of the ttc folder?

robertandremitchell (commented):
a few small comments. I think this broadly makes sense to me

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
nickclyde: Agreed, this and TTCMetadata could live in `shared_models`

nickclyde (commented):
Nice refactor! The Pydantic-driven serialization makes the lambda code much cleaner, and combining the duplicate `test_handler_success`/`test_handler_writes_outputs_to_s3` was a good call.

A few high-level thoughts:

- I think there's a behavior change for malformed eICRs with no relevant Schematron errors. I think reordering `EicrProcessor` construction (and especially the new unconditional `processor.eicr_metadata` access at the end of `_process_record_pipeline`) flips the passthrough reason for eICRs like `eicr_empty.xml` from `no_relevant_schematron_errors` to `ttc_exception`. I confirmed locally that `EicrProcessor(open('e2e/assets/eicr_empty/eicr_empty.xml').read()).eicr_metadata` raises `ValueError("No ID element found in eICR XML.")`, which is then caught by the handler's outer `except` and written as a TTC_EXCEPTION passthrough, matching the e2e snapshot change. I left an inline comment about a possible fix.

- The PR description says "reranker serializes score float to 3 decimal places" but the new snapshots show full-precision floats (e.g. `0.7127664685249329`). Was that change dropped?

- There may be reduced schematron-error coverage in the lambda tests. The new `packages/text-to-code-lambda/tests/assets/test_schematron_errors.xml` has 1 issue, whereas the previous shared asset had 4 (covering both `Lab Test Name Resulted` and `Lab Test Name Ordered`). The expectation constants `EXPECTED_RESULTED_ERRORS = 2` / `EXPECTED_ORDERED_ERRORS = 2` are no longer accurate (and one is now unused).

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
JNygaard-Skylight: I resolved this by removing the `raise` in the `EICRProcessor` when there is not an ID. In those cases we would still want to get the metadata, even if it's just the vendor. Right?

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
JNygaard-Skylight: Removing it seems to be safe. Good catch.

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
JNygaard-Skylight: You're probably right. I was hoping to distribute some of the logging to the other modules, as it feels like there's an abundance of them in this function. But there is a lot to be said about just keeping logging in one place.

packages/text-to-code-lambda/tests/test_lambda_function.py
JNygaard-Skylight: Weird the linter was not upset about that 🤷

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
nickclyde: Yeah I think that's right, thanks for addressing!

nickclyde (approved):
Thanks for addressing the feedback, great work Josh!

PR #575, opened 2026-05-21
style: remove force-single-line

### LINKED ISSUE #574: Bring Python rule inline with Other Dibbs projects.
In our `pyproject.toml`. we have this rule:
```
[tool.ruff.lint.isort]
force-single-line = true
```
This forces every Python import to be on a seperate line. This is out of line with other Dibbs projects who do not use this rule:
- [Refiner](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/pyproject.toml)
- [eCR Viewer](https://github.com/CDCgov/dibbs-ecr-viewer/blob/main/pyproject.toml)

`force-single-line` should be removed so that our formatting more closely aligns with other Dibbs projects.

JNygaard-Skylight:
## Description

In our `pyproject.toml`. we have this rule:
```
[tool.ruff.lint.isort]
force-single-line = true
```
This forces every Python import to be on a seperate line. This is out of line with other Dibbs projects who do not use this rule:
- [Refiner](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/pyproject.toml)
- [eCR Viewer](https://github.com/CDCgov/dibbs-ecr-viewer/blob/main/pyproject.toml)

`force-single-line` should be removed so that our formatting more closely aligns with other Dibbs projects.

## Related Issues

Closes #574


robertandremitchell (approved):
I'm okay with changing it especially if it matches other repos. I also find this one a bit more intuitive in grouping, especially when we have such nested imports

PR #577, opened 2026-05-22
Skip floating vX.Y / vX tags on APHL ECR push

nickclyde:
## Summary

- The v0.2.1 release workflow failed pushing to APHL ECR with `The image tag 'v0.2' already exists ... and cannot be overwritten because the tag is immutable`. APHL ECR has tag immutability enabled on every tag except `:latest`, so re-pointing the floating `vX.Y` and `vX` tags fails on every release after the first.
- Update the reusable build/push workflow to skip the `minor_tag` and `major_tag` values when iterating over the metadata-action tag list for the APHL push. GHCR still receives the full `vX.Y.Z` / `vX.Y` / `vX` / `latest` set.
- Update `docs/releases.md` and the `release.yml` header comment to reflect that APHL ECR only ever has `vX.Y.Z` and `latest` for each image.

Note: the v0.2.1 git tag was created and the GitHub Release was not (`release` job depends on `build`). After this lands we'll need to either dispatch the build manually or roll forward to v0.2.2 — happy to follow up with whichever approach we pick.

## Test plan

- [ ] CI green on this PR
- [ ] On the next release, confirm: GHCR has `vX.Y.Z`/`vX.Y`/`vX`/`latest`; APHL ECR has only `vX.Y.Z` and `latest`; no immutability error in the workflow logs

### DIFF
--- a/.github/workflows/_build-and-push.yml
+++ b/.github/workflows/_build-and-push.yml
@@ -126,20 +126,30 @@ jobs:
         id: aphl-ecr
         uses: aws-actions/amazon-ecr-login@v2
 
-      - name: Push every tag to APHL ECR
+      - name: Push tags to APHL ECR
         env:
           OWNER: ${{ steps.repo.outputs.owner }}
           NAME: ${{ matrix.name }}
           APHL_ECR: ${{ secrets.APHL_ECR_REPOSITORY_URL }}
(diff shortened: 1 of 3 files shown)


robertandremitchell (approved):
lgtm

JNygaard-Skylight (approved):
LGTM!

PR #578, opened 2026-05-26
refactor: Move some log statements closer to relevant code

JNygaard-Skylight:
## Description

There are  several log statements in the TTC lambda function that could be moved into their relevant modules:
- `"Evaluating candidates and selecting relevant text for each error in the eICR"` can be moved into the embedder
- `"Evaluating candidates and selecting relevant text"` can be moved into the evaluator.

## Related Issues

Closes # n/a

## Additional Notes

This was something I ended up doing for [PR 570](https://github.com/CDCgov/dibbs-text-to-code/pull/570) to reduce the size of the `_process_schematron_errors` function. However it can be easily moved into its own PR.


robertandremitchell (approved):
lgtm!

PR #589, opened 2026-05-29
feat: Add result cache to lambda loop

### LINKED ISSUE #533: Cache TTC pipeline results in a second OpenSearch index

bamader:
## Description

This PR adds the final integration of the result cache service into the per-invocation loop of the TTC lambda. CloudWatch metrics are also emitted as part of the loop, indicating whether the result cache was a hit or a miss.

## Related Issues

Closes #533 

## Additional Notes

The current implementation uses a lot of object creation to build properly typed data bundles and preserve our access patterns established elsewhere in the code. I don't love how this looks in most places (e.g. the result-cache `GET` API service, a lot of the tests, etc.) because there's some repeated, clunky-looking instantiation calls, but I went with this structure for two main reasons:

1. it preserves the way we access objects in other parts of the code base. For example, now both branches of the lambda invocation can treat variables the same, since whether or not the result-cache gets used, all objects and data fields have the expected information and expected types in them.
2. it mirrors the way other API responses are structured. For example, in `lambda_handler.py` > `retrieve_opensearch_results`, we go through the effort of turning the JSON response dictionary into a set of objects with expected typing and data structures. I assume this is because the OpenSearch GET response is actually quite dense (it's linked in the result cache function docs, but there are a lot of possible return fields), so explicitly defining this preserves logic for the rest of the code base. I adopted the same mentality here.

I'm not married to this class-based implementation, but I do think it's consistent with the rest of our code, so just wanted to outline the decision there.


bamader:
Unit tests are all passing, the end to end tests are what's left. I'm opening this up for review so that we can get eyes on any logic while I polish up any end to end failures.

pyproject.toml
```diff
@@ -95,10 +95,11 @@ select = [
   "W",    # pycodestyle warning
 ]
 ignore = [
-  "D100",   # undocumented-public-module
-  "E501",   # line-too-long - Handled by formatter
-  "RET504", # unnecessary-assign
-  "S314",   # recommend `defusedxml` over `xml` and `lxml`
+  "D100",    # undocumented-public-module
+  "E501",    # line-too-long - Handled by formatter
+  "PLR0915", # too many lines in function
```
robertandremitchell: I think this ought to be # no-qa exception the function rather than something we ignore. definitely in favor of trying to chunk/simplify functions as a broad principle, even if we have occasional exceptions

robertandremitchell (commented):
makes sense to me, awesome to see. I think you mentioned still adding some tests, so I'm happy to take a second look once those come through.

packages/text-to-code/src/text_to_code/services/result_cache.py
```diff
@@ -19,10 +25,62 @@ def get_cached_result(
       the TTC OpenSearch instance.
     :param index: The namespace of the Index to check.
     :param os_doc_id: The OpenSearch document `_id` parameter to check for.
+    :returns: The Source of the OpenSearch Cache Result, expressed as a structured
+      object.
     """
     response = opensearch_client.get(index=index, id=os_doc_id)
```
nickclyde: I believe `opensearch_client.get()` raises [opensearchpy.exceptions.NotFoundError](https://opensearch-project.github.io/opensearch-py/api-ref/exceptions.html#opensearchpy.NotFoundError) (404) when the document doesn't exist, rather than returning a body with `"found": False`. If that's right, then on a cache miss this raises instead of returning `None`, and the exception propagates to the per-record handler and sends the eICR to TTC_EXCEPTION passthrough. I think we need to pass `ignore=404` (e.g. `opensearch_client.get(index=index, id=os_doc_id, ignore=404)`) in order keep relying on the `found` check.

packages/text-to-code/src/text_to_code/services/result_cache.py
```diff
@@ -19,10 +25,62 @@ def get_cached_result(
...
     if response and response["found"]:
-        return response["source"]
+        response_source = response["source"]
```
nickclyde: I think the OpenSearch GET response returns the document body under the `_source` key (see [here](https://github.com/opensearch-project/opensearch-py/blob/main/guides/search.md?plain=1#L301)), so `response["source"]` would raise a `KeyError` against a real cluster.

packages/text-to-code/src/text_to_code/services/result_cache.py
```diff
@@ -61,9 +125,11 @@ def put_new_cached_result(  # noqa: PLR0913
...
         text=candidate_input,
         data_field=data_field,
-        loinc_code=json.dumps(loinc_code.__dict__),
+        loinc_code=loinc_code,
         search_score=search_score,
         reranker_score=reranker_score,
+        opensearch_retrieved_scores=opensearch_retrieved_scores,
+        reranker_processed_results={"results": reranker_processed_results},
     )
     put_response = opensearch_client.index(index=index, id=cache_key, body=new_cache_hit)
```
nickclyde: I believe passing the pydantic model directly as `body` raises a `SerializationError` at runtime, since opensearchpy's JSON serializer doesn't know how to serialize a pydantic `BaseModel` (I confirmed this locally with opensearchpy 3.0). Because `.index()` is mocked in the tests, this path isn't exercised.

There's also a round-trip concern: `get_cached_result` reads the nested OpenSearch result back using the alias keys (`_shards`, `_index`, `_id`, `_score`, `_source`), but those only appear if we serialize with `by_alias=True`. Could we store `body=new_cache_hit.model_dump(by_alias=True)` so the serialization works and the stored keys match what the GET path expects?

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -306,6 +312,30 @@ def _save_ttc_outputs(
...
+    """Helper method for creating a cache-result metric dict.
+
+    Output is generated using Embedded Metric Format Specification.
+    :param hit_value: A string indicating whether the metric denotes a cache hit
+      or a cache miss.
+    :return: A dictionary in EMFS that can be emitted by the Logger.
+    """
+    return {
+        "_aws": {
+            "Timestamp": int(datetime.timestamp(datetime.now(UTC))),
```
nickclyde: I think the EMF `_aws.Timestamp` field needs to be in epoch milliseconds, but `int(datetime.timestamp(...))` produces seconds- maybe we could multiply by 1000?

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -306,6 +312,30 @@ def _save_ttc_outputs(
...
+            "CloudWatchMetrics": [
+                {
+                    "Namespace": "ttc-lambda-cache-metrics",
+                    "Dimensions": [["lastLoincUpdate"]],
+                    "Metrics": [{"Name": "result_cache_value_status"}],
+                }
+            ],
+        },
+        "lastLoincUpdate": "02-23-2026",
+        "result_cache_value_status": hit_value,
```
nickclyde: Looking at the [docs](https://awscli.amazonaws.com/v2/documentation/api/2.3.2/reference/cloudwatch/put-metric-data.html), I think CloudWatch metric values have to be numeric. Should we do something like 1 for hit, 0 for miss? Or alternatively we could use two separate metrics.

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -306,6 +312,30 @@ def _save_ttc_outputs(
...
+            "Timestamp": int(datetime.timestamp(datetime.now(UTC))),
+            "CloudWatchMetrics": [
+                {
+                    "Namespace": "ttc-lambda-cache-metrics",
+                    "Dimensions": [["lastLoincUpdate"]],
+                    "Metrics": [{"Name": "result_cache_value_status"}],
+                }
+            ],
+        },
+        "lastLoincUpdate": "02-23-2026",
```
nickclyde: Would it be worth reading this from an env var? I'm not sure about hardcoding it. Any thoughts @BradySkylight?

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -362,56 +392,104 @@ def _process_record_pipeline(
...
-                                new_translation=new_translation,
-                            ),
-                        )
-                    else:
-                        unmatched_message = "Reranker did not return any results."
+                # We've got a hit--no need to run our usual processes, we'll
+                # just pull out the fields we want to use directly
+                if cached_result is not None:
+                    cache_hit_metric = _get_cache_metric("hit")
+                    logger.info(cache_hit_metric)
```
nickclyde: Should we be using the powertools `Metrics` utility here instead of logger.info? https://docs.aws.amazon.com/powertools/python/latest/core/metrics

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -362,56 +392,104 @@ def _process_record_pipeline(
...
-                    "Querying OpenSearch with the relevant text strings and retrieving code suggestions",
-                    status="processing",
-                )
-                query = QueryBuilder().with_vector_search(vector_parameters).build()
-
-                opensearch_retrieved_scores = lambda_handler.retrieve_opensearch_results(
-                    query=query, index=OPENSEARCH_INDEX, opensearch_client=opensearch_client
+                # Before we run the full embedding, searching, and reranking process,
+                # first let's check if we've seen this nonstandard input before
+                cache_key = sha256(
```
nickclyde: This `cache_key` derivation is duplicated inside `put_new_cached_result`, which recomputes the same sha256 over `candidate_input.strip().lower() + "|" + data_field`. If one side ever changes, the keys would diverge. Could we extract a small shared helper (e.g. `compute_cache_key(text, data_field)`) used by both?

packages/text-to-code-lambda/tests/test_lambda_function.py
```diff
@@ -74,6 +86,111 @@ def test_handler_success(
...
+                data_field="Lab Test Name Ordered",
+                loinc_code=Code(
+                    code="82041-5",
+                    code_system="2.16.840.1.113883.6.1",
+                    code_system_name="LOINC",
+                    display_name="Weed Allerg Mix3 IgE Qn",
+                    original_text="A custom code in display name.",
+                ),
+                search_score=0.88,
+                reranker_score=7127664685249329,
```
nickclyde: Tiny nit: should this be `0.7127664685249329`?

packages/text-to-code-lambda/tests/test_lambda_function.py
```diff
@@ -354,6 +474,8 @@ def test_handler_adds_unmatched_error_when_selected_candidate_has_no_opensearch_
         assert reranker_mock.call_count == 0
 
         ttc_output = _get_serialized_object(f"{TTC_OUTPUT_PREFIX}{mock_aws_setup.persistence_id}")
+        print("TTC OUTPUT")
+        print(ttc_output)
```
nickclyde: These look like leftover debugging prints. Could we remove them?

pyproject.toml
robertandremitchell: updated

e2e/snapshots/test_e2e/test_upload_and_process/eicr_covid/eicr_covid_augmented_eicr.xml
robertandremitchell, line 1: looking at this file...not sure why the comments on augmentation fell out, gonna double check

e2e/snapshots/test_e2e/test_upload_and_process/eicr_test/eicr_test_augmentation_metadata.json
```diff
@@ -1,23 +1,8 @@
...
-      },
-      "new_translation_xpath": "/ClinicalDocument/component/structuredBody/component/section/entry/observation/code/translation",
-      "schematron_error_xpath": "/ClinicalDocument/component[1]/structuredBody[1]/component[1]/section[1]/entry[1]/observation[1]"
-    }
-  ],
-  "original_eicr_id": "c8516bdc-8bb2-40aa-8dae-20a77546488f",
-  "passthrough": false,
-  "passthrough_reason": null
+  "nonstandard_codes": [],
+  "original_eicr_id": "2025/09/03/1-5f84c7a5-91d7f5c6a2b7c9e08f0d1234",
```
robertandremitchell: gotta figure out what's happening here,  this would should be the original, not the augmented eicr id

nickclyde (commented):
Thanks for getting the result cache wired into the lambdas @bamader!  The biggest issues I'm seeing:

- Cache reads on a miss likely raise. `get_cached_result` relies on `opensearch_client.get(...)` returning `{"found": False}`, but I believe `.get()` raises `NotFoundError` on a 404. Since misses are the common case at rollout, an unhandled raise here would push eICRs into TTC_EXCEPTION passthrough.
- Cache writes likely raise. `put_new_cached_result` passes a pydantic model as the OpenSearch `body`, which opensearchpy can't serialize (I confirmed a `SerializationError` locally with opensearchpy 3.0). It also needs `by_alias=True` so the stored keys round-trip with the GET path.
- The cache hit/miss metrics probably don't emit. There are three separate issues (the `_aws` block nests under `message`, the metric value is a string, and the timestamp is in seconds) detailed inline.
- Testing gap. All OpenSearch interactions (`.get()`, `.index()`) are mocked, and the mocks use shapes that don't match real OpenSearch behavior (eg `"source"` instead of `_source`). That's why the issues above pass CI. An integration-style test against a   Docker container OpenSearch round-trip, or at least mocks that mirror the real response/serialization contract, would catch these. Since this PR closes #533, it would be worth confirming the end-to-end path against a live index before merge.

Another minor nit, in `packages/text-to-code/src/text_to_code/models/result_cache.py`, `cached_at` uses `default=datetime.now(UTC).isoformat()`, which is evaluated once at import time, so every cache entry without an explicit timestamp gets the time at module load. A `default_factory=lambda: datetime.now(UTC).isoformat()` would stamp each entry at creation.

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
BradySkylight: I just saw this.  I don't like the hardcoding either.  Maybe an env or a config?  Do we need this loinc_date?  Could we pull it from the file name perhaps?

JNygaard-Skylight: Could this be left as a follow-up ticket? Seems like there's a couple of different things we could do that are not strictly related to this PR.

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
JNygaard-Skylight: Added an intEnum where hit==1 and miss=0

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
JNygaard-Skylight: I'm not super confident but I think I'm using the metrics utility correctly.

packages/text-to-code/src/text_to_code/services/result_cache.py
JNygaard-Skylight: I think I resolved this.

packages/text-to-code/src/text_to_code/services/result_cache.py
```diff
@@ -19,10 +24,62 @@ def get_cached_result(
       the TTC OpenSearch instance.
     :param index: The namespace of the Index to check.
     :param os_doc_id: The OpenSearch document `_id` parameter to check for.
+    :returns: The Source of the OpenSearch Cache Result, expressed as a structured
+      object.
     """
     response = opensearch_client.get(index=index, id=os_doc_id)
```
nickclyde: Following up on the earlier comment: I think `opensearch_client.get()` raises `opensearchpy.exceptions.NotFoundError` (404) on a cache miss rather than returning a body with `found: False`. Since the result cache starts empty, most inputs would be misses, and the raised exception propagates up to the per-record handler (`lambda_function.py` lines 80-87), which catches it and routes the eICR to the `TTC_EXCEPTION` passthrough instead of running the normal TTC pipeline. Could we pass `ignore=404` (e.g. `opensearch_client.get(index=index, id=os_doc_id, ignore=404)`) so we can keep relying on the `found` check? The tests mock `.get()`, so this path isn't exercised by the suite.

packages/text-to-code/src/text_to_code/services/result_cache.py
```diff
@@ -51,19 +110,23 @@ def put_new_cached_result(  # noqa: PLR0913
...
         text=candidate_input,
         data_field=data_field,
-        loinc_code=json.dumps(loinc_code.__dict__),
+        loinc_code=loinc_code,
         search_score=search_score,
         reranker_score=reranker_score,
+        opensearch_retrieved_scores=opensearch_retrieved_scores,
+        reranker_processed_results={"results": reranker_processed_results},
     )
     put_response = opensearch_client.index(index=index, id=cache_key, body=new_cache_hit)
```
nickclyde: I think this earlier thread was marked resolved, but the code still passes the raw pydantic model as `body`. I confirmed locally with opensearchpy 3.0.0 that `JSONSerializer.dumps()` raises `SerializationError` on a `BaseModel`, so `.index(..., body=new_cache_hit)` would raise at runtime. Also: `get_cached_result` reads the nested result back using alias keys (`_shards`, `_index`, `_id`, `_score`, `_source`), which only appear if we serialize with `by_alias=True`. I think we can resolve this with `body=new_cache_hit.model_dump(by_alias=True)` so serialization works and the stored keys match what the GET path expects. Since `.index()` is mocked in the tests, this isn't caught by the suite either.

nickclyde (commented):
Thanks for addressing most of the previous comments, @JNygaard-Skylight! Two that I noticed were still outstanding, and may lead to the code not working as written. It's probably worth deploying this branch to our AWS and giving it a real test before merging - maybe we could all do a quick mob sesh on it?

BradySkylight (approved):
Nice!

nickclyde (approved):

PR #590, opened 2026-06-01
Add Embedding Vectors to Embedding Records for LOINC Updates

### LINKED ISSUE #454: Create process to take valueset delta updates and generate embeddings

BradySkylight:
## Description

Added functionality in the LOINC Update process to create actual embedding vectors, using our same 'embed' service in TTC.  This also handles the functionality for creating insert (id that is numeric) vs. update records (id of "").  This also handles some general clean-up and re-org of the code (adding new folder for changelog and moving some functions into different files as appropriate).

**NOTE**: The diagram included here is for how this will work in AWS not necessarily how this PR is functioning, but this is a step closer to making this a reality!

## Related Issues

Closes #454 

## Additional Notes
Note that in another PR this will all be changed and moved to run in AWS instead of locally using S3 buckets and logging statements.  See diagram below:

SNOINC_UPDATE_PROCESS

## Checklist for Reviewers


robertandremitchell (approved):
lgtm!

packages/data-curation/src/data_curation/terminologies/maintenance.py
```diff
@@ -27,30 +35,29 @@ def update_loinc_embeddings():
...
-    # 
-    # This will be part of the creating the actual embeddings work
-    #  It will have to succeed from this step to making the embeddings
-    #   into a file or files and then we can update/remove the existing
-    # csv file
-    # Handled in Ticket #454
+    # add embeddings to any of the records for the various descriptions
+    # if there are none, no looping will occur    
+    # TODO: In Subsequent PR update this to be a logging statement
+    print(f"LOINC Lab Name Embedding Records to add: {len(loinc_updates)}")
```
nickclyde: I think this can raise a `NameError` when there are no updates. In the `else` branch above (the "No updates found" case) `loinc_updates` is never assigned, and the previous version of this function had a `return` there that's now been removed. So when `file_date > loinc_version_date`, execution falls through to here and `len(loinc_updates)` references an unbound variable. Could we initialize `loinc_updates = []` before the `if`, or restore an early `return` in the `else` branch?

packages/data-curation/src/data_curation/terminologies/loinc.py
```diff
@@ -420,36 +470,45 @@ def get_loinc_embedding_records(current_loinc_dict: dict, new_version: str) -> l
...
+            if (update_loinc_record["short_name"]
+                and current_loinc_record["short_name"].strip() != update_loinc_record["short_name"].strip()
+            ):
+                change_log["Changes"]["short_name"] += 1
                 changes.append("short_name")
-            if current_loinc_record["long_name"] != update_loinc_record["long_name"]:
-                change_log["long_name"] += 1
+            if (current_loinc_record["long_name"]
+                and current_loinc_record["long_name"].strip() != current_loinc_record["long_name"].strip()
+            ):
```
nickclyde, line 490: I believe there's a copy-paste bug here: this compares `current_loinc_record["long_name"]` against itself, so the condition is always `False` and `long_name` changes will never be detected or counted. The other branches show the intended shape, so I think this probably wants:

```python
if (update_loinc_record["long_name"]
    and current_loinc_record["long_name"].strip() != update_loinc_record["long_name"].strip()
):
```

(switching the guard to `update_loinc_record` as well, to match the surrounding branches.)

packages/data-curation/src/data_curation/terminologies/loinc.py
```diff
@@ -399,19 +434,34 @@ def get_loinc_embedding_records(current_loinc_dict: dict, new_version: str) -> l
...
+        :returns: List of embedding records (dictionaries).
     """
     delta_extract_rows = get_loinc_lab_names(new_version)
+    # get the max number to ensure no id collisions in Opensearch
+    # by getting the max loinc codes in the current file *5 for all the
+    # different 'names/text' that will be used to create embeddings
+    #  then add 1
+    loinc_record_max_id = (len(current_loinc_dict)*5)+1
     embedding_records = []
+    change_log_filename = f"{LAB_NAMES}_DELTA_{datetime.now().strftime('%Y%m%d')}.csv"
```
nickclyde: Small thing: this change log is written as JSON (`write_change_log_file` -> `save_json_file` uses `json.dump`), but the filename ends in `.csv`. Should this be `.json` so the extension matches the contents?

packages/data-curation/src/data_curation/terminologies/maintenance.py
```diff
@@ -27,30 +35,29 @@ def update_loinc_embeddings():
...
+    for loinc_update_record in loinc_updates:
+        if (loinc_update_record.get("description") is not None
+            and loinc_update_record.get("description").strip() is not None
```
nickclyde: I think `.strip()` always returns a string, so `.strip() is not None` is always `True` and this second clause is effectively a no-op. If the intent is to skip blank descriptions, I think it could just be `and loinc_update_record.get("description").strip()`

packages/data-curation/src/data_curation/terminologies/general.py
```diff
@@ -113,3 +115,116 @@ def load_extract_file_to_dict(filename: str) -> list[dict]:
...
+        :param append_to_file: Boolean - Do you want to overwrite the file or append?
+            Default is set to False, so overwrite the file.
+
+        :returns: Nothing.
+    """
+    if not filename.strip():
+        print("No filename supplied.  Failed to save CSV file!")
+        return
+
+    if contents is None and len(contents) == 0:
```
nickclyde: I think this guard may not behave as intended: with `and`, if `contents is None` then `len(contents)` still gets evaluated and raises `TypeError` on `None`, while a non-`None` empty list never trips the check because the first clause is already `False`. I think it should be `if contents is None or len(contents) == 0:`? The same pattern appears in `save_json_file` (line 184).

packages/data-curation/src/data_curation/terminologies/maintenance.py
```diff
@@ -1,11 +1,19 @@
...
                                                      LAB_NAMES,
-                                                     get_loinc_embedding_records)
+                                                     get_loinc_embedding_records,
+                                                     extract_full_loinc_lab_names)
 from data_curation.terminologies.general import (get_latest_extract_file_name, 
                                                        get_date_from_filename, 
-                                                       load_extract_file_to_dict)
+                                                       load_extract_file_to_dict,
+                                                       save_jsonl_file)
+from text_to_code.services.embedder import embed
```
nickclyde, line 11: Just flagging: `data-curation`'s `pyproject.toml` doesn't declare `text-to-code` (or `utils`) as a dependency, so this import relies on the workspace already having those packages installed. It also has an import-time side effect, since `embedder.py` instantiates `SentenceTransformer(TTC_RETRIEVER)` at module load, so importing this module loads the model. Probably fine for a local script, but it might be worth adding `text-to-code` to the dependencies so it's explicit.

nickclyde (commented):
Thanks for this, Brady! The reorg looks great. Pulling the `save_*` helpers out of `terminology_extraction.py` and into `general.py`, and moving the `extract_full_loinc_*` functions next to the other LOINC code, both make sense. The id-generation tests are a nice add too.

I left a few inline comments. The two I'd most want a look at before merge are:
- `long_name` change detection compares the current record against itself, so those changes are never picked up.
- `update_loinc_embeddings()` can hit a `NameError` on the no-updates path now that the `else` `return` was removed.

A few non-blocking design/coverage questions for the broader feature (these are partly aimed at the AWS rework you mentioned, so feel free to defer):

- Name changes create new inserts rather than updating. For an existing LOINC whose `short_name`/`long_name`/etc. changed (not a `LOINC Type` change), `_create_embedding_records` assigns a brand-new numeric id, so the record is treated as an insert. That leaves the old embedding (old text + old id) in OpenSearch alongside the new one. Are we planning to add new logic in a future PR to lookup the old id and use that instead?

- `id` value for updates. `LOINC Type` changes set `id` to `""` (empty string), while the PR description mentions "id of 0" for update records. Just confirming `""` is what the index loader expects to mean "update existing".

- Off-by-one in the starting id. `loinc_record_max_id = (len(current_loinc_dict)*5)+1`, but `_create_embedding_records` does `new_id += 1` before assigning the first id, so the value `len*5+1` is never actually used. No collision results, just a one ID gap at the start. 

- The new `_create_embedding_records` tests are great, but there's no test around `get_loinc_embedding_records`, which is exactly where the `long_name` bug lives. A small test there would catch this class of issue. Also `test_loinc.py` is missing a trailing newline.

BradySkylight:
> 
> * `long_name` change detection compares the current record against itself, so those changes are never picked up.
> * `update_loinc_embeddings()` can hit a `NameError` on the no-updates path now that the `else` `return` was removed.

Good finds... both of these are fixed!


> A few non-blocking design/coverage questions for the broader feature (these are partly aimed at the AWS rework you mentioned, so feel free to defer):
> 
> * Name changes create new inserts rather than updating. For an existing LOINC whose `short_name`/`long_name`/etc. changed (not a `LOINC Type` change), `_create_embedding_records` assigns a brand-new numeric id, so the record is treated as an insert. That leaves the old embedding (old text + old id) in OpenSearch alongside the new one. Are we planning to add new logic in a future PR to lookup the old id and use that instead?

This is intended.  We do not want to update the 'Name' for an existing embedding, as we are creating a totally new embedding for the 'new' name for this loinc code.  That was what we wanted.  Additive vs. updating.  The only time we update an embedding record is if the type changes because that will modify the 'filter' of the query results.

packages/data-curation/src/data_curation/terminologies/maintenance.py
BradySkylight: Nice find...this may resolve some of the issues I was running into.

packages/data-curation/src/data_curation/terminologies/general.py
BradySkylight: Yeah.. that was a total lame miss on my part.  Fixed!

packages/data-curation/src/data_curation/terminologies/maintenance.py
BradySkylight: Good catch.  Fixed!

packages/data-curation/src/data_curation/terminologies/loinc.py
BradySkylight: Yes!  Fixed.

packages/data-curation/src/data_curation/terminologies/loinc.py
BradySkylight: Fixed

packages/data-curation/src/data_curation/terminologies/maintenance.py
BradySkylight: Fixed.

PR #593, opened 2026-06-01
SPIKE: Embedding variability

nickclyde:
## Description

Adds a self-contained research spike under `docs/spikes/embedding-variability/` that investigates a concern raised about TTC: the same clinical text produces slightly different vector embeddings depending on the compute environment (CPU vs GPU, Azure vs AWS vs local Mac). Because TTC relies on representing English meaning as numbers, and we fine-tune/evaluate on Azure GPUs but run in production on AWS CPUs, we wanted to know whether that difference is large enough to change the medical codes the system returns.

The spike re-embeds the same ~500 LOINC display-name strings in **five environments** (Azure GPU, Azure CPU, Mac CPU, Mac GPU/MPS, and the production AWS Lambda runtime), then compares the vectors and runs real KNN queries against the production OpenSearch index to see whether retrieval results change.

**Findings — the compute environment is not a meaningful source of variability for the retriever:**

- **Vectors are practically identical.** Azure GPU vs AWS CPU embeddings of the same string differ by ~`6.5e-06` in L2 distance and ~`2.3e-11` in cosine distance — the floating-point precision floor. Semantically they mean the same thing.
- **Retrieval is stable.** The top-1 retrieved LOINC code is **identical across all five environments for all 500 strings.** The rest of the top-10 candidate list (its membership, ordering, and scores) is stable to within the search engine's *own* re-query noise — i.e. the approximate (HNSW) index returns slightly different deep-rank orderings even when you re-query the exact same vector, and the cross-environment variation is no larger than that.
- **The one historical edge case is a tie, not a CPU/GPU effect.** A description that flipped its top-1 in an earlier run turns out to have two candidates with *identical* scores; which one ranks first is decided arbitrarily by index internals.

**Recommendation: no action.** We do not need to re-embed the index on CPU, and a CPU/GPU vector swap does not need to be raised with APHL. The reranker stage (which does its own internal embedding) was intentionally left out of scope and is flagged as an open item, bounded by the fact that the retriever is the dominant, trusted component and upcoming confidence/margin thresholds will further limit reranker exposure.

The deliverable is `report.html` — a single self-contained page organized around the four questions the data-science review posed. Other outputs are gitignored; the findings and full reproduction steps are captured in `docs/spikes/embedding-variability/README.md`.

## Additional Notes

- **No production code is touched.** Everything lives under `docs/spikes/embedding-variability/`, which is excluded from lint/type checks.
- The analysis hits the production OpenSearch index read-only (KNN queries) and reads the existing index vectors; it writes nothing back.
- Scope/limitations are documented in the report and README: the reranker stage (Q3) and an index-swap control (Q4) are deliberately not measured, the analysis is on the 500-string subset, and the determinism baseline is included so reviewers can distinguish embedding drift from the approximate-search noise floor.


bamader (approved):
Looks great! Really glad we went through these steps, and this most recent report really addresses all the concerns I had.

BradySkylight (approved):
This is insanely detailed.  Nice!

PR #594, opened 2026-06-01
Add Schematron validation to aws_e2e.sh

nickclyde:
## Summary

The AWS e2e scripts used to upload a hand-written, static schematron-error file at the start of the run, so the pipeline was always driven by canned input and never actually checked that augmentation fixed anything. This wires real Schematron validation into both scripts: validate the generated eICR up front to produce the report the TTC lambda consumes, then re-validate the augmented eICR at the end and fail if any errors remain.

Validation runs in-process via the existing `validation` package (Saxon + the APHL Text-to-Code schematron).

## What changed

- **`validation` package** - new `build_schematron_report_xml()` plus a `python -m validation` CLI (`report` / `check`). The serializer captures the SVRL message + test, normalizes Saxon EQName XPaths (`/Q{...}ClinicalDocument[1]/...`) to the plain form the TTC reader expects, and emits the NIST `<Report>` shape the pipeline already parses. `validate_eicr()`'s public behavior is unchanged.
- **`scripts/aws_e2e.sh`** - generates the real report from the templated eICR, then re-validates the augmented output and hard-fails (red banner, non-zero exit) if errors remain.
- **`scripts/batch_aws_test.sh`** - same per case; a case passes only if the predicted LOINC matches *and* validation is clean; the run exits non-zero if any case fails.
- **`Dockerfile.augmentation`** - installs the `validation` package and pre-bakes its XSLT artifacts at build time. The augmentation lambda already imported `validation` (for its own post-augmentation check) but the image never bundled it, so it crashed at import with `No module named 'validation'`; pre-baking also avoids writing the XSLT cache on the read-only Lambda filesystem at runtime.
- Removed the now-unused `scripts/test_schematron_errors.xml`; updated `scripts/README.md` and added validation-package tests.

## Testing

- `validation` unit tests, full unit suite (341 passed), and the moto e2e suite (7 passed) all green; ruff + ty clean.
- Built `Dockerfile.augmentation` for `linux/amd64` and confirmed inside the container: the handler module imports, artifacts are pre-baked, and `validate_eicr` runs against a read-only filesystem.
- Ran the full deployed pipeline end to end. `aws_e2e.sh` completed with the "✓ Validation clean — Schematron errors resolved" banner.

## Deployment note

The augmentation-lambda fix only takes effect once the `ttc-augmentation-lambda` image is rebuilt and redeployed from this branch.


robertandremitchell (commented):
running into some issues running the scripts locally, can we maybe see a demo during parking lot?

code otherwise looks great, makes sense

packages/validation/src/validation/main.py
```diff
@@ -34,21 +41,62 @@ class ValidationResult(FrozenBaseModel):
...
+    """Compile the schematron and run it against the eICR, returning failed asserts.
+
+    :param eicr: The eICR XML document as a string.
+    :param redo_all_steps: When True, regenerate the cached XSLT artifacts.
+    :return: The list of failed asserts emitted by the validator.
+    """
     logger.info("Starting eICR Validation")
     logger.info(f"For eICR: {eicr}")
-    errors: list[ValidationResult] = []
+    asserts: list[_RawAssert] = []
```
BradySkylight, line 85: Nice!

packages/validation/src/validation/main.py
```diff
@@ -96,14 +148,86 @@ def validate_eicr(eicr: str | None = None, redo_all_steps: bool = False) -> list
...
+    and ``issue/test`` children, matching the message text against known error
+    enums.
+
+    :param eicr: The eICR XML document as a string.
+    :param redo_all_steps: When True, regenerate the cached XSLT artifacts.
+    :return: The schematron validation report as a NIST ``<Report>`` XML string.
+    """
+    results = "\n".join(
+        _validation_result_xml(raw) for raw in _run_validator(eicr, redo_all_steps=redo_all_steps)
+    )
```
BradySkylight, line 222: This is pretty sweet

BradySkylight (approved):
LGTM!  This is great to incorporate our schematron validation into the process!

robertandremitchell (approved):
lgtm!

packages/validation/tests/test_validation.py
```diff
@@ -179,3 +195,63 @@ def test_validation_raises_when_validator_errors(
...
+
+    report = build_schematron_report_xml(eicr)
+
+    assert "Q{" not in report  # Saxon EQName prefixes must be normalized away
+
+    issues = list(ET.fromstring(report).iter("issue"))
+    assert len(issues) == 1
+    assert issues[0].findtext("message") == FAKE_MESSAGE
+    context = issues[0].findtext("context")
+    assert context is not None
```
JNygaard-Skylight, line 255: This is redundant as the below asserts would also fail if context was `None`.

PR #598, opened 2026-06-02
feat!: Remove configuration from Augmentation

### LINKED ISSUE #595: Remove Configuration from Augmentation
## Summary
The Augmentation package was created to potentially be used by multiple different projects. To that end A seperate `config.py` module was made with a config model. However, now that the augmentation package will only be used internally this can removed. We no longer need to be able to pass in a configuration model.

## Acceptance Criteria
- [ ] Remove the ability to pass in a configuration
- [ ] Remove the config module in the Augmentation package. However, it may still make sense to keep the various constants, such as xpaths in a single location. 
- [ ] Any of refactors and simplifications that can be made

JNygaard-Skylight:
## Description

Removes the configuration model and the ability to pass in a configuration to the augmenter

## Related Issues

Closes #595 

## Additional Notes

For reasons I do not understand this caused the Sample 9 to fail the e2e test, but coping the the `voc_ttc.xml` into the validators output directory resolves it. 🤷‍♂️


BradySkylight (approved):
This looks great!  Glad to see Augmentation become more simple!  🚀

PR #607, opened 2026-06-05
refactor: ruff data curation

### LINKED ISSUE #572: Enable Ruff in data curation package
Ruff should be abled in the data curation package. It is currently excluded so that when Ruff was added the initial PR would not be too large.

## pyproject.toml
```
[tool.ruff]
# Wholesale ignoring several directories as I do not want to have to resolve everything in one PR,
# but I would highly recommend enabling directories as we work on them, and enabling/disabling
# rules at a more granular level as needed.
exclude = [
  ".git",
  ".pytest_cache",
  "__pycache__",
  "azure_scripts",
  "data",
  "data-curation",
  "docs",
  "model-tuning",
]
```

JNygaard-Skylight:
## Description

- Resolves ruff issues in the data curation package.
- Deletes the archived directory in data curation.
- Moves data emulation code into `data_emulation` package.

## Related Issues

Closes #572


packages/data-curation/pyproject.toml
```diff
@@ -20,3 +20,17 @@ build-backend = "hatchling.build"
 
 [tool.hatch.build.targets.wheel]
 packages = ["src/data_curation"]
+
+
+[tool.ruff]
+extend = "../../pyproject.toml"
+extend-exclude = ["src/data_curation/data_emulation"]
+
+[tool.ruff.lint]
```
robertandremitchell, line 29: do we need a follow-up ticket to make this info logging? may be a broader question about where/how to log things for this part of the work

robertandremitchell (commented):
broadly looks good! looks like a few small coverage gaps opened up with some of the reorg that should be addressed. I have a broader question to better under how we might approach logging v. printing in this particular section of code, but I don't think that has to be solved in this ticket.

packages/data-curation/tests/test_post_process.py
```diff
@@ -93,91 +108,116 @@ def test_no_modality(self):
...
+            truncated
+            == "Myelin associated glycoprotein/Sulfated glucuronic paragloboside IgM Ab [Titer] in Se"
+        )
+
 
 class TestDetermineEligiblePostProcessingOptions:
-    all_options = [
-        "poc", "modality", "delimiter", "truncation", "syntax", "pound", "deletion", "dot"
-    ]
+    def __init__(self) -> None:
```
nickclyde: I think this change disables the three tests in this class. pytest can't collect a test class that defines an `__init__` constructor, so it spits out `PytestCollectionWarning: cannot collect test class 'TestDetermineEligiblePostProcessingOptions' because it has a __init__ constructor` and skips `test_ordinary_lcn`, `test_code_with_pounds_and_delimiters`, and `test_code_with_dots_and_truncation` (I confirmed this locally). Was `__init__` was introduced to satisfy RUF012 (mutable class attribute)? Could we instead keep it as a class attribute annotated with `ClassVar`, e.g. `all_options: ClassVar[list[str]] = [...]` (with `from typing import ClassVar`), or move it to a module-level constant? That lets those three tests run.

packages/data-curation/src/data_curation/terminologies/http_client.py
```diff
@@ -0,0 +1,17 @@
+from typing import Any
+
+import requests
```
nickclyde, line 3: Naming this module `requests.py` while it also does `import requests` works here because the inner import is absolute, but I think it feels a bit weird. If anyone later adds a plain `import requests` elsewhere in this `terminologies/` package, they'd pick up this local module instead of the library. Could we rename it to something like `http_client.py` to avoid the shadowing?

packages/data-curation/src/data_curation/terminologies/loinc.py
```diff
@@ -281,35 +271,33 @@ def get_loinc_umls_urls(loinc_results, loinc_rows_list):
     return loinc_rows_list
 
 
-def process_loincs_for_umls_urls() -> dict:
-    """Process that constructs and makes the various API calls necessary to gather
-        the UMLS data about the current LOINC codes.  The lower functions, related
-        to ATOM/UMLS process are leveraged to extract and organize the data together
-        for the enhancement/other LOINC files.
+def process_loincs_for_umls_urls() -> dict | None:
```
nickclyde: Since this function now raises `ValueError` when the result isn't a dict (rather than ever returning `None`), I think the `-> dict | None` annotation is no longer accurate and could just be `-> dict`. Worth noting too that this is a behavior change: when the LOINC API errors during UMLS processing, `_process_loinc_valueset` now returns `[]`, which trips the isinstance check and raises here, whereas previously it returned `None`. That seems reasonable for an error path, just flagging it.

pyproject.toml
```diff
@@ -92,8 +91,11 @@ ignore = [
 ]
 
 [tool.ruff.lint.flake8-annotations]
-# Turn off return type annotations for __init__
+# Relax typing requirements for certain situations
+allow-star-arg-any = true
 mypy-init-return = true
+suppress-dummy-args = true
+suppress-none-returning = true
```
nickclyde, line 98: These flake8-annotations relaxations apply to every package in the workspace right? Not just data-curation. `suppress-none-returning` means functions that only return `None` no longer need an explicit `-> None`, which loosens the "annotations required on public functions" convention repo-wide (and it looks like the `-> None` got dropped from `__init__` in the index-lambda test). Was the intent to relax this globally, or would it be better to scope these to the data-curation package's own `[tool.ruff]` config?

packages/data-curation/src/data_curation/terminologies/vsac.py
```diff
@@ -21,62 +14,50 @@
...
-    """Function to get the all VSAC SNOMED Codes and terms for 
-        Diagnosis/Problems via the NLM/NIH VSAC API. There is an 
-        underlying function that processes the results from various
-        VSAC API results into a common structure that is leveraged 
-        within this function.
-
-        :returns: A list of dictionaries containing SNOMED Problem records
-            including codes and text.
+    """Function to get the all VSAC SNOMED Codes and terms for Diagnosis/Problems via the NLM/NIH VSAC API. There is an underlying function that processes the results from various VSAC API results into a common structure that is leveraged within this function.
+
```
BradySkylight, line 37: Is this the format we are striving for, no returns in a paragraph?  I was following a different pattern for a long time.  I'm ok either way, just want to make sure I know how to docstring properly.  :)

packages/data-curation/src/data_curation/terminologies/http_client.py
```diff
@@ -0,0 +1,17 @@
...
+from requests.models import Response
+
+_TIMEOUT = 60
+
+STATUS_CODE_OK = 200
+
+
+def get_with_timeout(
```
BradySkylight, line 11: Great addition to help simplify things more.  Nice!

nickclyde (commented):
Thanks Josh, looks good! Just a few notes:

packages/data-curation/pyproject.toml
JNygaard-Skylight: Print statements seem perfectly acceptable for a script like this, but I don't see a reason to not use logging.

packages/data-curation/src/data_curation/terminologies/vsac.py
JNygaard-Skylight: I don't have much of a preference in how or where to add line breaks in in docstrings. I personally don't mind letting the IDE wrap long lines so I do not have to worry about it. However, specifically for the first line of docstrings, the convention is for it to be one line.

pyproject.toml
JNygaard-Skylight: My intent was for this to be repo-wide. These seem like reasonable relaxations, but I don't feel strongly about it, and it is easy enough to remove them if we would rather be consistent with typing.

nickclyde (approved):
Thanks for addressing the comments, great work Josh!

PR #610, opened 2026-06-09
Refactoring augmentation-lambda persistence id assignment

robertandremitchell:
## Description

This PR is intended to try to clean up and streamline _process_record so that the logic for when and how we fall back to original EICR cleaner. I personally find the currently written process a bit tough to follow and am hoping this will help clarify.

## Additional Notes

## Checklist

## Checklist for Reviewers


e2e/snapshots/test_e2e/test_upload_and_process/sample9/sample9_augmentation_metadata.json
```diff
@@ -1,8 +1,8 @@
 {
-  "augmented_eicr_id": "2025/09/03/1-5f84c7a5-91d7f5c6a2b7c9e08f0d1234",
+  "augmented_eicr_id": "1.2.840.114350.1.13.370.2.7.8.688883.608381234",
```
JNygaard-Skylight, line 2: If the augmented eICR ID is changing, shouldn't we also see it change in the augmented eICR snapshot?

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -173,125 +107,178 @@ def _process_record(record: SQSRecord) -> None:
...
+        )
+
+    return output
+
+
+def _build_original_eicr_output(  # noqa: PLR0913
+    persistence_id: str,
+    original_eicr_id: str,
+    original_eicr: str,
+    nonstandard_codes: list[NonstandardCodeInstanceMetadata],
```
JNygaard-Skylight: An empty string is passed in a couple of places. I think it would be a little cleaner if it defaulted to `None` and in the this function we could do something like `if nonstandard_codes is None: nonstandard_codes = []`.

robertandremitchell: makes sense. Updated!

e2e/snapshots/test_e2e/test_upload_and_process/sample9/sample9_augmentation_metadata.json
robertandremitchell: I'm not sure I follow? This is a passthrough case, where because there's no relevant schematron errors we don't actually generate a new eICR, the xml in the snapshot folder is a 1:1 carryover of the original eICR in the asset folder. So I think the logic on this is right, where we wouldn't want that passed through eICR to have a different ID.

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -173,125 +107,179 @@ def _process_record(record: SQSRecord) -> None:
...
+    original_eicr_id = persistence_id
+
+    try:
+        augmenter = EICRAugmenter(
+            document=original_eicr,
+            nonstandard_codes=augmenter_input.nonstandard_codes,
+            deterministic_id_seed=augmenter_input.persistence_id,
+        )
+
+        original_eicr_id = str(augmenter.original_eicr_id)
```
nickclyde: Constructing the `EICRAugmenter` before the passthrough check is a significant change. Previously a TTC passthrough never touched the augmenter at all. Now every passthrough record parses the XML and deep-copies the entire document tree. Since passthrough (`no_relevant_schematron_errors`) is presumably the most common case in production, that is a fair amount of extra work per record.

If the goal is just to get the real document ID for passthrough metadata, could we consider a lighter-weight ID extraction (e.g. a small helper that parses and reads the `id/@root` attribute) instead of building the full augmenter with its deepcopy?

e2e/snapshots/test_e2e/test_upload_and_process/eicr_empty/eicr_empty_augmentation_metadata.json
```diff
@@ -1,6 +1,6 @@
 {
   "augmented_eicr_id": "2025/09/03/1-5f84c7a5-91d7f5c6a2b7c9e08f0d1234",
-  "error": null,
+  "error": "Unable to find tag in eICR document for XPath: /ClinicalDocument/id/@root",
```
nickclyde: Do you know why this changed? Before it was a passthrough with no error. I don't think this will effect downstream consumers like RCKMS but I was curious.

packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
```diff
@@ -245,6 +264,13 @@ def test_handler_writes_original_eicr_when_augmentation_fails(
             object_key=f"{AUGMENTATION_METADATA_PREFIX}{TEST_PERSISTENCE_ID}",
         )
         metadata = json.loads(metadata_raw)
+        assert metadata["original_eicr_id"] == EXPECTED_ORIGINAL_EICR_ID
+        assert metadata["augmented_eicr_id"] == EXPECTED_ORIGINAL_EICR_ID
+        assert metadata["original_eicr_id"] != TEST_PERSISTENCE_ID
+        assert metadata["augmented_eicr_id"] != TEST_PERSISTENCE_ID
+        assert metadata["error"] == "augmentation boom"
+        assert metadata["passthrough"] is True
+        assert metadata["passthrough_reason"] == PassthroughReason.AUGMENTATION_EXCEPTION
```
nickclyde, line 287: These assertions are a nice strengthening of the test. One gap I noticed: this test patches `EICRAugmenter` with a mock whose `augment()` raises, so `original_eicr_id` has already been set to the real document ID by the time the exception fires. The other failure branch, where the `EICRAugmenter` constructor itself raises and `original_eicr_id` falls back to `persistence_id` (line 142 in lambda_function.py), is only covered indirectly by the `eicr_empty` e2e snapshot. Could we add a unit test for that path too, asserting the metadata IDs fall back to the persistence ID?

JNygaard-Skylight (dismissed):
Approved!

nickclyde (commented):
Thanks Rob! Just a couple thoughts:

e2e/snapshots/test_e2e/test_upload_and_process/eicr_empty/eicr_empty_augmentation_metadata.json
robertandremitchell: This changed based on the error read added for the try/except logic in augmentation-lambda. In theory, we only hit this if they send us an empty/malformed XML. We could add error messaging to the ttc-lambda instead, since in this case, we basically are getting an empty/unreadable eicr xml, so I don't know if we want it to flow to augmentation if we can flag it earlier in the workflow.

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
robertandremitchell: That makes sense to me. I think when I was tooling around with it earlier, I had just a small XML parser for this step, but i didn't want to have duplicative processing code.

I was wondering if it just makes sense to extend `TTCAugmenterInput` to include ID, since we do end up needing to read the ID more often anyway? Is there any concern with going down that road?

packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
robertandremitchell: updated!

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
robertandremitchell: Decided to try it, and I think this ended up looking cleaner? curious your thoughts

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -173,125 +107,173 @@ def _process_record(record: SQSRecord) -> None:
...
+        logger.info(
+            "Augmentation processing completed",
+            status="passthrough" if output.metadata.passthrough else "success",
+            passthrough_reason=output.metadata.passthrough_reason,
+        )
 
-            original_eicr_id = str(augmenter.original_eicr_id)
 
-            metadata = augmenter.augment()
+def _build_augmentation_output(
```
JNygaard-Skylight, line 124: Would it be cleaner if there was only 1 or 2 returns and the logic was used to determine the inputs to the `_build_original_eicr_output`/`_build_augmented_eicr_output`?

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
```diff
@@ -173,125 +107,173 @@ def _process_record(record: SQSRecord) -> None:
...
+        )
+
+    return output
+
+
+def _build_original_eicr_output(  # noqa: PLR0913
+    persistence_id: str,
+    original_eicr_id: str,
+    original_eicr: str,
+    passthrough_reason: PassthroughReason | None,
```
JNygaard-Skylight, line 224: Would there ever be reason why we would create the original eICR without a passthrough reason?

I think this is currently needed because the `TTCAugmenterInput` class is written such that it is technically possible that `passthrough` could be `True` but `passthrough_reason` could be `None`, or vice versa. I do not think there is a reason why we need a seperate `passthrough` boolean when we could test if `passthrough_reason` is set.

This does not need to happen in this PR, I'll create a ticket.

packages/shared-models/src/shared_models/__init__.py
```diff
@@ -69,6 +69,7 @@ class TTCAugmenterInput(FrozenBaseModel):
     """Input for the augmentation service."""
 
     persistence_id: str
+    original_eicr_id: str | None = None
```
JNygaard-Skylight: We should be using the `CdaInstanceIdentifier` here because it is possible for eICRs to have the same root but different extensions.

e2e/snapshots/test_e2e/test_upload_and_process/eicr_empty/eicr_empty_augmentation_metadata.json
robertandremitchell: with the latest update, this should now be moot, as the TTCAugmenterInput includes the eicr_id, so that if there is no eicr_id, augmentation never reaches this point to generate an error message.

e2e/snapshots/test_e2e/test_upload_and_process/augmentation_metadata.json
JNygaard-Skylight, line 1: Where these output files left over from a previous version of tests?

robertandremitchell: yeah, this should've been deleted in a prior PR where we reorganized the sample eicrs into respective assets folders.

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
robertandremitchell: yeah, that's a good question. I think that could be a good follow-up ticket to just rely on passthrough_reason.

packages/shared-models/src/shared_models/__init__.py
robertandremitchell: updated!

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
robertandremitchell: I changed up this section somewhat, but I had trouble originally making sense of the code when it was 1-2 returns versus trying to capture the different variations that lead to building the original eicr output. it probably could be simplified but I worry a little about losing some of the transparency into what is happening for each error

packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
nickclyde: Agreed, looks good now!

nickclyde (approved):
Thanks Rob!!

PR #619, opened 2026-06-10
Add synchronous demo API and browser frontend for Text-to-Code

nickclyde:
## Summary

Adds a simple static frontend (HTML/CSS/JS) that lets anyone map lab-test strings to LOINC codes interactively. Also takes a CSV of lab test strings, sends them to a small synchronous API, and shows TTC's suggested codes for review and annotated export. The API reuses the existing embed → KNN → rerank pipeline, so the demo exercises the real matching logic against the real OpenSearch domain.

image

### Synchronous service (`packages/text-to-code-lambda`)

- **`service.py`** — the core. `code_for_text()` mirrors the inner matching logic of `_process_record_pipeline` (embed the text, run the KNN query, rerank, return the top hit as a `Code`), but takes a plain string instead of an eICR candidate. `results_for_inputs()` batches inputs while preserving order and skips blank rows
- **`api_handler.py`** — an AWS Lambda **Function URL** handler intended to deploy as a second Lambda built from the **same container image** as the SQS batch worker, with the `CMD` overridden to point at this handler. Handles CORS preflight, validates the request body, and caps requests at 100 inputs.
- **`local_server.py`** — a FastAPI dev server exposing the same logic for local development against the real OpenSearch domain ("local backend → AWS data"). `fastapi`/`uvicorn` are dev-group dependencies only, so they stay out of the Lambda image.

Both entry points call into `service.py`, so the matching logic lives in exactly one place.

### Request / response

```text
POST /text-to-code
{ "inputs": ["Glucose measurement", ...], "data_field": "Lab Test Name Ordered" }
```

```json
{ "results": [
  { "input": "Glucose measurement", "matched": true, "code": "110283-9",
    "code_system": "2.16.840.1.113883.6.1", "code_system_name": "LOINC",
    "display_name": "Glucose [Measurement]" }
] }
```

`data_field` is optional and defaults to `Lab Test Name Ordered`; `Lab Test Name Resulted` switches the OpenSearch `loinc_type` terms filter. Unrecognized values fall back to the default with a warning log.

### Demo frontend (`frontend/`)

A small static page styled to follow USWDS tokens and patterns, split into three files under `frontend/`:

- **`index.html`** — page structure / markup
- **`styles.css`** — the USWDS-token stylesheet
- **`app.js`** — behavior (gov banner, nav, the demo try-it flow, and the CSV upload → review → export flow, plus the API calls)

`app.js` loads as a classic script at the end of `<body>`, so the inline event handlers in the markup still resolve and load timing is unchanged. Features:

- Single-string try-it box plus CSV upload (first column is read as inputs, optional header row dropped)
- Results table with each LOINC code linked to its `loinc.org` page
- Per-row reviewer annotation and "Export annotated CSV" download

### Local runner (`frontend/demo.sh` + `frontend/.env.example`)

`./frontend/demo.sh` loads a `.env` file, starts the FastAPI backend on port 8080 and a static server for the `frontend/` page on port 8000, and tears both down on Ctrl+C. `frontend/.env.example` documents the required `OPENSEARCH_ENDPOINT_URL`, AWS credential options, and optional knobs (index name, region, ports, CORS origin, `HF_TOKEN`).

## Testing

- New unit tests for `service.py` cover: reranker order winning over OpenSearch order, no-hits and empty-rerank short circuits, blank-input skipping, and `data_field` parsing/fallback. The embedder is stubbed so tests don't load the retriever model.
- Verified end-to-end locally via `./frontend/demo.sh` against the deployed OpenSearch domain.

## Notes for reviewers

- Nothing in the production batch path changes — `lambda_function.py` is untouched, and the new dependencies are dev-only.
- The Function URL Lambda (`api_handler.py`) is code-only in this PR; Terraform for deploying it will come in a later PR.
- `frontend/app.js` hardcodes `API_BASE = http://127.0.0.1:8080` for the local dev server; `demo.sh` warns if you change `BACKEND_PORT` without updating it.


index.html
JNygaard-Skylight, line 1: Could the CSS live in its own file?

robertandremitchell (commented):
Broadly seems fine. I was able to follow the setup instructions to get started. I can't recall where we landed on whether the CSV functionality worked, but it could be good to have that saved somewhere, like the `scripts` folder or an `assets` folder in the ttc-lambda or something. I just gave it a dummy CSV to see what that page looks like.

I have some concerns about maintaining the `index.html` file, from both a design and code standpoint in terms of keeping continuity with other dibbs products. Do we know if this is a like a step 0 to drawing up an epic to build this more fully? Or should we plan to focus on the maintainability of what was output here (i.e., Josh's note about separate css file, breaking up the code in other ways)

m-goggins:
+1 to Rob and Josh's comments. Is this step 0? (don't we already have step 0 which is the frontend we deleted from the repo)? I'm a little worried about maintainability as is...don't want to re-create the wheel just to delete it.

index.html
nickclyde: Definitely, agree that would be cleaner, probably the js could be extracted too

nickclyde:
Good points team. Let's bring these concerns to Ashley on backlog refinement Monday.

packages/text-to-code-lambda/src/text_to_code_lambda/api_handler.py
JNygaard-Skylight, line 1: I think it would make sense if the `api_handler.py`, `local_server.py`, and `service.py` where in their own module to make it clear that they are related and somewhat separated from the typical lambda handler service.

JNygaard-Skylight (approved):
Approved!

JNygaard-Skylight (changes requested):
The top level readme should be updated. It still have content for the old frontend and API.

JNygaard-Skylight (approved):
Approved (again)!

robertandremitchell (approved):

PR #627, opened 2026-06-12
Add Hugging Face Model Info to TTC Metadata

### LINKED ISSUE #615: Add Hugging Face Details to TTC metadata

m-goggins:
## Description

This PR adds Hugging Face model info to the TTC Metadata output. For each model, the output now includes:

- id: `NCHS/ttc-reranker-mvp` or `NCHS/ttc-retriever-mvp`
- author: `NCHS`
- created_at: timestamp of model creation
- last_modified: timestamp of model modification

To get this information, we have to hit the Hugging Face API 1x for each model at start up using the new function `get_model_info`. We could also get information from the [Model Card](https://huggingface.co/docs/hub/model-cards),  which wouldn't require accessing the API, but it doesn't include the date information and I _believe_ requires us to update the model card when we add a new model to Hugging Face (so not as automatic). I thought `date_modified` would be important for keeping track of which model was used but we could also go by name from the Model Card. Happy to hear others' thoughts!

## Related Issues

Closes #615


packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -130,6 +126,10 @@ def _write_ttc_exception_passthrough_output(record: SQSRecord, error: Exception)
             passthrough=True,
             passthrough_reason=PassthroughReason.TTC_EXCEPTION,
             error=str(error),
+            model_info=TTCModelInfo(
+                reranker=_RERANKER_MODEL_INFO,
+                retriever=_RERANKER_MODEL_INFO,
```
robertandremitchell: I think this is meant to be `_RETRIEVER_MODEL_INFO` here and below, which should update the snapshots.

robertandremitchell (changes requested):
looks good to me! one typo to fix

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
m-goggins: Ahh good catch! That would have been super annoying to find later

packages/text-to-code/src/text_to_code/models/model_info.py
```diff
@@ -0,0 +1,19 @@
+from datetime import datetime
+
+from pydantic import BaseModel
+
+
+class ModelInfo(BaseModel):
```
JNygaard-Skylight, line 6: This does not need to be mutable, or accept additional attributes, right? If that is the case it should use `FrozenModel`.

packages/text-to-code/src/text_to_code/models/model_info.py
```diff
@@ -0,0 +1,19 @@
...
+    """Info about a Text-to-Code model sourced from Hugging Face."""
+
+    id: str | None
+    author: str | None
+    created_at: datetime | None
+    last_modified: datetime | None
+
+
+class TTCModelInfo(BaseModel):
```
JNygaard-Skylight, line 17: Same question regarding `BaseModel`/`FrozenModel`.

packages/text-to-code/src/text_to_code/services/embedder.py
```diff
@@ -4,10 +4,12 @@
 from torch import Tensor
 
 from text_to_code.models.registry import TTC_RETRIEVER
+from text_to_code.services.utils import get_model_info
 
 logger = logging.getLogger(__name__)
 
 _RETRIEVER = SentenceTransformer(TTC_RETRIEVER)
+_RETRIEVER_MODEL_INFO = get_model_info(TTC_RETRIEVER)
```
JNygaard-Skylight: What is the purpose of this constant?

packages/text-to-code/src/text_to_code/services/reranker.py
```diff
@@ -2,8 +2,10 @@
 from typing_extensions import TypedDict
 
 from text_to_code.models.registry import TTC_RERANKER
+from text_to_code.services.utils import get_model_info
 
 _RERANKER = CrossEncoder(TTC_RERANKER)
+_RERANKER_MODEL_INFO = get_model_info(TTC_RERANKER)
```
JNygaard-Skylight: What is the purpose of this constant?

packages/text-to-code/src/text_to_code/models/model_info.py
m-goggins: Oh good call!

packages/text-to-code/src/text_to_code/services/embedder.py
m-goggins: It contains the model info for the retriever. Since we have to call the HF API to get this and we likely only need to get it once per start up, I thought it made sense to store it like this so that we're not fetching it per lambda invocation but I'm open to other suggestions.

packages/text-to-code-lambda/tests/test_lambda_function.py
```diff
@@ -72,7 +72,8 @@ def test_handler_success(
         ttc_metadata_output = _get_serialized_object(
             f"{TTC_METADATA_PREFIX}{mock_aws_setup.persistence_id.removesuffix('.xml')}.json"
         )
-        snapshot.assert_match(ttc_metadata_output, "handler_success_ttc_metadata_output.json")
+        print(ttc_metadata_output)
```
robertandremitchell: should this be back to a snapshot?

packages/text-to-code/src/text_to_code/models/model_info.py
```diff
@@ -0,0 +1,23 @@
+from datetime import datetime
+
+from pydantic import BaseModel, ConfigDict
+
+
+class ModelInfo(BaseModel):
```
robertandremitchell, line 6: should this be frozenbasemodel? looks like that comment got resolved

robertandremitchell (approved):
lgtm, few small nits but otherwise good to go

packages/text-to-code/src/text_to_code/models/model_info.py
m-goggins: I believe I made both of the models frozen using the [ConfigDict](https://pydantic.com.cn/en/concepts/models/#rootmodel-and-custom-root-types). Let me know if there's a better way to do this though!

packages/text-to-code/src/text_to_code/services/embedder.py
JNygaard-Skylight: The `_` should be removed as it is not intended to to be a private variable.

PR #634, opened 2026-06-17
Bumping python to allow 3.13

robertandremitchell:
## Description

We just merged a dependabot flagged lambda from 3.12 to 3.13, which broke some of our images: https://github.com/CDCgov/dibbs-text-to-code/actions/runs/27693398902/job/81911551823

If we are concerned with allowing this, we can always rollback that PR for the time being since it is perhaps a bigger breaking change than I realized at the time.

The `torch` being pinned to 2.9.1 was causing issues with a universal change to the primary `pyproject.toml` to allow 3.13 wheels. We have an open PR about pushing up that version that may resolve that issue, but as noted in that dependabot ticket, there could be an issue with an upgrade there: https://github.com/CDCgov/dibbs-text-to-code/pull/621

May be worth considering at the same time. Per eng sync convo, we should push to at least 3.13 and also pin back the Dockerfiles to just be 3.13.

I added a stop to keep any 3.14 out, but 🤷🏽  I think I would be okay with just >=3.13, especially given we will probably want to review/change versions of our dockerfile images to 3.14 more deliberately.

## Related Issues

Related to https://github.com/CDCgov/dibbs-text-to-code/pull/633
Semi-related to https://github.com/CDCgov/dibbs-text-to-code/pull/621

## Additional Notes

## Checklist

## Checklist for Reviewers


JNygaard-Skylight (dismissed):
The readmes say `Python 3.13 or higher`, but the `pyproject.toml`s say `requires-python = ">=3.13, <3.14"`. For simplicity's sake I think we should pin to 3.13, and not make any guarantees about future versions of Python.

robertandremitchell:
> The readmes say `Python 3.13 or higher`, but the `pyproject.toml`s say `requires-python = ">=3.13, <3.14"`. For simplicity's sake I think we should pin to 3.13, and not make any guarantees about future versions of Python.

updated!

nickclyde (approved):

PR #636, opened 2026-06-17
test: ttc metadata in e2e snapshot

JNygaard-Skylight:
## Description

Create snapshots for the ttc metadata as part of the local e2e tests.


robertandremitchell (approved):
looks good to me. looks like the snapshots are failing for not yet reflecting the removal of the `passthrough` boolean from metadata.

Could be worth creating a follow-up ticket about the decimal changes in the e2e snapshots. Not sure if we need that precision in e2e tests since we've seen those shift before so that we're not always having to update them.

e2e/snapshots/test_e2e/test_upload_and_process/eicr_covid/eicr_covid_ttc_metadata.json
```diff
@@ -24,7 +24,6 @@
       "last_modified": "2026-03-30T14:12:37Z"
     }
   },
-  "passthrough": false,
   "passthrough_reason": null,
```
BradySkylight, line 27: This make more sense.

BradySkylight (approved):
🚀

PR #637, opened 2026-06-18
Increasing data curation coverage

robertandremitchell:
## Description

A few small coverage gaps on `data-curation/`. I also noticed a small place in `validation/` tests where we should be using the constant.

I had a local issue with validation; I had to remove and regenerate the .tmp files to get tests to work locally. This doesn't impact github actions since it regenerates the tests each time. I added a small cleanup ticket (https://app.zenhub.com/workspaces/dibbs-text-to-code-68756bf8ea06bb00319391c0/issues/gh/cdcgov/dibbs-text-to-code/638) to see if there's anything we can add locally to either regenerate the files with each run, or a check to determine if it should, but it could be overkill.

## Additional Notes

## Checklist

## Checklist for Reviewers

### DIFF
--- a/packages/data-curation/tests/test_loinc.py
+++ b/packages/data-curation/tests/test_loinc.py
@@ -99,6 +99,52 @@ def test_extract_full_loinc_lab_names(monkeypatch: pytest.MonkeyPatch) -> None:
     assert saved_files[0][2] is False
 
 
+def test_extract_full_loinc_lab_names_handles_empty_api_results(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+    capsys: pytest.CaptureFixture[str],
+) -> None:
+    payload: dict[str, object] = {
+        "ResponseSummary": {
+            "RecordsFound": 0,
+            "RowsReturned": 1,
+            "Next": None,
+        },
+        "Results": [],
+    }
+    consumer_names_file = tmp_path / "consumer_names.csv"
+    consumer_names_file.write_text(
+        "LoincNumber|ConsumerName\n",
+        encoding="utf-8",
+    )
+    saved_files: list[tuple[str, list[loinc.LoincRow], bool]] = []
+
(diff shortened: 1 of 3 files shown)


packages/data-curation/tests/test_loinc.py
```diff
@@ -191,6 +191,18 @@ def mock_get_loinc_consumer_names(loinc_rows: list[dict[str, str]]) -> list[dict
     assert calls["api_url"] == loinc.LOINC_BASE_URL + f"query={loinc.LOINC_LAB_NAMES_QUERY}"
 
 
+def test_get_loinc_lab_names_raises_when_valueset_is_not_list(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    def mock_process_loinc_valueset(api_url: str, loinc_vs_type: str) -> dict[str, str]:
+        return {"unexpected": "dict"}
+
+    monkeypatch.setattr(loinc, "_process_loinc_valueset", mock_process_loinc_valueset)
```
JNygaard-Skylight: As a general rule we should not be testing private methods. Regardless, I think we could avoid this coverage issue by removing all of the type checking in loinc.py if `_process_loinc_valueset` always returned a list. See [this PR](https://github.com/CDCgov/dibbs-text-to-code/pull/639).

JNygaard-Skylight (changes requested):
Thank you for writing tests for the http client! 

I'm pretty we could avoid having to do all of these tests for when the valueset is not a list if we just changed `_process_loinc_valueset` to always return a list.  tried that in [this PR](https://github.com/CDCgov/dibbs-text-to-code/pull/639) and it seems to work.

packages/data-curation/tests/test_loinc.py
robertandremitchell: makes sense. updated!

JNygaard-Skylight (approved):
Approved!

PR #640, opened 2026-06-18
chore: Configure release notes

JNygaard-Skylight:
## Description

Updates out `release.yaml` so that it is more in line with other Dibbs projects. [ecr-refiner](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/.github/release.yml), [ecr-viewer](https://github.com/CDCgov/dibbs-ecr-viewer/blob/main/.github/release.yaml).

## Related Issues
n/a

## Additional information
This just tells Github what to include/exclude in the autogenerated release notes and where to put them, based on their labels. This will create two sections: "🏕 Features", and "👒 Dependencies".  Features include all PRs that do not have the "dependencies", "chore", or "tech debt" labels. Dependencies include all PRs with the "dependencies" label.

During the last eng sync we talked a little about [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), which are just a standardized way to include some additional information in your commit string. While it is possible with actions to use conventional commits to drive release note generation, this PR just uses the builtin Github note generation which is driven by labels.


m-goggins (approved):
Looks good!

PR #641, opened 2026-06-18
fix: augmenter passes through only on newly-introduced validation errors

nickclyde:
## Problem

After augmenting an eICR, the augmentation lambda re-validated the result and treated **any** schematron finding as fatal, passing through the original (un-augmented) eICR. This blocked good augmentations whenever the input carried unrelated, pre-existing data-quality issues, most commonly lab **result values** that lack an interpretation code (`ttc-labResultValue-PQNoInterp` / `-STNoInterp`), which the augmenter never touches (it only adds lab-test-**name** `<translation>` elements).

In the latest APHL training rerun, persistence ID `2026/06/11/19d4812b…` was fully augmented (9 LOINC translations) but then discarded via `passthrough_reason: augmentation_validation_failure`, so RCKMS never received an augmented document. The blocking errors (28× `PQNoInterp`, 3× `STNoInterp`) were already present in the input eICR. Pre/post-anonymization validation both report an identical 48 result observations / 17 interpretation codes, so the gate would also block these in production, independent of anonymization.

## Fix

Switch to a **diff-based gate** (`_introduced_validation_error`): validate the original eICR as a baseline and pass through only when augmentation **introduces new** `(error_id, location)` findings. Pre-existing findings cancel out of the diff. A clean augmented result short-circuits the baseline call, so the common path still runs a single `validate_eicr`. Conservative passthrough on validation exceptions is preserved. The recorded `error` now lists only the newly-introduced findings.

## Testing

- `ruff` + `ty` clean.
- Unit (`augmentation-lambda`): updated the existing gate test to diff semantics and added `test_handler_emits_augmented_eicr_when_validation_errors_are_preexisting`.
- e2e: the `sample7` (`nullFlavorResultValues`) fixture correctly flips from passthrough to augmented; snapshots regenerated (`passthrough: true->false`, `error` cleared, translations added, only the pre-existing finding remains). The success-path assertion now checks "no new findings vs. original" rather than "zero findings".
- Reproduced on the real `19d4812b` eICR: old gate - passthrough (31 errors present); new gate emits the augmented eICR, augmentation fixing 8 `noCode` errors and introducing 0.

Bumps version to **v0.2.3** for release to APHL.

### DIFF
--- a/e2e/snapshots/test_e2e/test_upload_and_process/sample7/sample7_validation_results.json
+++ b/e2e/snapshots/test_e2e/test_upload_and_process/sample7/sample7_validation_results.json
@@ -1,9 +1,5 @@
 [
   {
-    "error_id": "ttc-labTestNameResulted-code-missing",
-    "location": "/Q{urn:hl7-org:v3}ClinicalDocument[1]/Q{urn:hl7-org:v3}component[1]/Q{urn:hl7-org:v3}structuredBody[1]/Q{urn:hl7-org:v3}component[8]/Q{urn:hl7-org:v3}section[1]/Q{urn:hl7-org:v3}entry[1]/Q{urn:hl7-org:v3}organizer[1]/Q{urn:hl7-org:v3}component[1]/Q{urn:hl7-org:v3}observation[1]"
-  },
-  {
     "error_id": "ttc-labResultValue-noCode",
     "location": "/Q{urn:hl7-org:v3}ClinicalDocument[1]/Q{urn:hl7-org:v3}component[1]/Q{urn:hl7-org:v3}structuredBody[1]/Q{urn:hl7-org:v3}component[8]/Q{urn:hl7-org:v3}section[1]/Q{urn:hl7-org:v3}entry[1]/Q{urn:hl7-org:v3}organizer[1]/Q{urn:hl7-org:v3}component[1]/Q{urn:hl7-org:v3}observation[1]"
   }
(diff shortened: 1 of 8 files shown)


m-goggins (approved):
This looks good to me! 

One question that I have that is probably more for APHL. How should eICRs that pass through TTC but contains errors beyond TTC (and the augmenter) be handled in RCKMS? I don't want APHL to come back to us and say that TTC augmentation passed on eICRs with errors. Perhaps it's just about communicating this change (in writing) to them.

robertandremitchell (approved):
makes sense to me

e2e/snapshots/test_e2e/test_upload_and_process/sample7/sample7_validation_results.json
```diff
@@ -1,8 +1,4 @@
 [
-  {
-    "error_id": "ttc-labTestNameResulted-code-missing",
-    "location": "/Q{urn:hl7-org:v3}ClinicalDocument[1]/Q{urn:hl7-org:v3}component[1]/Q{urn:hl7-org:v3}structuredBody[1]/Q{urn:hl7-org:v3}component[8]/Q{urn:hl7-org:v3}section[1]/Q{urn:hl7-org:v3}entry[1]/Q{urn:hl7-org:v3}organizer[1]/Q{urn:hl7-org:v3}component[1]/Q{urn:hl7-org:v3}observation[1]"
-  },
   {
     "error_id": "ttc-labResultValue-noCode",
```
BradySkylight, line 3: I think this error id is incorrect actually.  See the list of proper error ids here:  https://docs.google.com/document/d/1P3FVz3nul6bRYZOj1jPWkBXM8voJeKC2NyUP-wUGY5I/edit?tab=t.0#heading=h.17zp4vjriomr

nickclyde: Will address separately!

PR #652, opened 2026-06-22
fix: removing passthrough from metadata and switching to passthrough_reason

### LINKED ISSUE #628: Remove `passthrough` attribute from `TTCAugmenterInput`
## Summary
The `TTCAugmenterInput` currently looks like:
```
class TTCAugmenterInput(FrozenBaseModel):

    persistence_id: str
    original_eicr_id: str | None = None
    nonstandard_codes: list[NonstandardCodeInstance] = []
    passthrough: bool = False
    passthrough_reason: PassthroughReason | None = None
```
This erroneously would allow for a situation where `passthrough == True` but `passthrough_reason is None` or the inverse: `passthrough == False` and `passthrough_reason is not None`. 

The class could be simplified by removing `passthrough` and using `passthrough_reason is not None` instead of `passthrough == True`

robertandremitchell:
## Description

Removes `passthrough` from code and metadata output. `passthrough_reason` being not null as the method to determine whether the eICR is passthrough.

## Related Issues

Closes #628 

## Additional Notes

## Checklist

## Checklist for Reviewers

### DIFF
--- a/packages/augmentation-lambda/tests/snapshots/test_augmentation_lambda_function/test_handler_writes_original_eicr_when_ttc_output_has_passthrough_reason/ttc_passthrough.json
+++ b/packages/augmentation-lambda/tests/snapshots/test_augmentation_lambda_function/test_handler_writes_original_eicr_when_ttc_output_has_passthrough_reason/ttc_passthrough.json
@@ -3,6 +3,5 @@
   "error": null,
   "nonstandard_codes": [],
   "original_eicr_id": "c8516bdc-8bb2-40aa-8dae-20a77546488f",
-  "passthrough": true,
-  "passthrough_reason": null
+  "passthrough_reason": "no_code_matches"
 }
--- a/packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
+++ b/packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
@@ -184,11 +184,6 @@ class TestHandler:
     def test_handler_writes_augmented_eicr_when_ttc_output_passthrough_reason_is_missing(
         self, example_sqs_event, mock_aws_setup, mock_lambda_context
     ) -> None:
-        original_eicr = lambda_handler.get_file_content_from_s3(
-            bucket_name=S3_BUCKET,
-            object_key=f"TextToCodeSubmissionV2/{TEST_PERSISTENCE_ID}",
-        )
-
         ttc_output = TTCAugmenterInput.model_validate(
             json.loads(
                 lambda_handler.get_file_content_from_s3(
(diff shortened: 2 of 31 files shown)


packages/augmentation-lambda/tests/snapshots/test_augmentation_lambda_function/test_handler_writes_original_eicr_when_ttc_output_has_passthrough_reason/ttc_passthrough.json
```diff
@@ -3,6 +3,5 @@
   "error": null,
   "nonstandard_codes": [],
   "original_eicr_id": "c8516bdc-8bb2-40aa-8dae-20a77546488f",
-  "passthrough": true,
-  "passthrough_reason": null
+  "passthrough_reason": "no_code_matches"
```
m-goggins, line 6: For my own understanding, is this the example the original ticket was referring to? Where `passthrough` was true but `passthrough_reason` was null. I'm tryign to understand why `passthrough_reason` would change to "no_code_matches" here

m-goggins (approved):
Looks good!  Just one question for my own understanding.

packages/augmentation-lambda/tests/snapshots/test_augmentation_lambda_function/test_handler_writes_original_eicr_when_ttc_output_has_passthrough_reason/ttc_passthrough.json
robertandremitchell: The new snapshot has "no_code_matches" because this test is not covering the invalid passthrough=true/passthrough_reason=null case. The original test was meant to try to cover what should happen in the event the two signals differed (roll back to original eICR)

packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
```diff
@@ -215,36 +211,28 @@ def test_handler_writes_original_eicr_when_ttc_output_passthrough_reason_is_miss
...
         assert result["message"] == "Augmentation processed successfully!"
         assert result["num_success_eicrs"] == 1
-        augmenter_mock.assert_not_called()
 
         augmented_eicr = lambda_handler.get_file_content_from_s3(
             bucket_name=S3_BUCKET,
             object_key=f"{AUGMENTED_EICR_PREFIX}{TEST_PERSISTENCE_ID}",
         )
-        assert augmented_eicr == original_eicr
+        assert augmented_eicr != original_eicr
```
JNygaard-Skylight: This might be beyond the scope of this PR, but I wonder if it would be worth asserting something like the IDs are different. Otherwise a stray whitespace character would satisfy this.

JNygaard-Skylight (approved):
This looks good to me! I had just one question, but I don't think it it quite related to this PR, unless you want to look at it.

packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
robertandremitchell: yeah, that makes sense to me. I don't love this assertion either, so I can tinker with it.

PR #659, opened 2026-06-23
fix: handle optional fields

### LINKED ISSUE #657: Augmentation fails when it can not find optional attributes
setId
VersionNumber

JNygaard-Skylight:
## Description

`setId` and `versionNumber` are optional fields in an eICR. However augmentation will crash if it does not find them. This PR changes the methods for getting an element by XPath so that there is a specific one for required and optional elements so that the absence of an element can be handled.

## Related Issues

Closes #657


packages/augmentation/src/augmentation/services/eicr_augmenter.py
```diff
@@ -89,26 +91,42 @@ def augment(self) -> AugmentResult:
...
-        old_set_id = self._get_augmented_tag_by_xpath("/ClinicalDocument/setId")
-        self._add_previous_element_comment("new-document-setId ", old_set_id)
-        self._replace_element("/ClinicalDocument/setId", self._get_new_set_id())
+        old_set_id = self._get_element_by_xpath(self._augmented_element, "/ClinicalDocument/setId")
+        new_set_id = self._get_new_set_id()
+        if old_set_id is not None:
+            self._add_previous_element_comment("new-document-setId ", old_set_id)
+            self._replace_element("/ClinicalDocument/setId", new_set_id)
+        else:
+            self._augmented_element.append(new_set_id)
```
nickclyde: Does this append to the very end of `ClinicalDocument` (after `<component>`)? I think `setId`/`versionNumber` are header elements, right? Should they go right after `languageCode`?

packages/augmentation/src/augmentation/services/eicr_augmenter.py
```diff
@@ -89,26 +91,42 @@ def augment(self) -> AugmentResult:
...
-        old_version = self._get_augmented_tag_by_xpath("/ClinicalDocument/versionNumber")
-        self._add_previous_element_comment("new-document-versionNumber ", old_version)
-        self._replace_element("/ClinicalDocument/versionNumber", self._get_new_version_number())
+        old_version = self._get_element_by_xpath(
+            self._augmented_element, "/ClinicalDocument/versionNumber"
+        )
+        new_version = self._get_new_version_number()
+        if old_version is not None:
+            self._add_previous_element_comment("new-document-versionNumber ", old_version)
+            self._replace_element("/ClinicalDocument/versionNumber", self._get_new_version_number())
```
nickclyde: Could we pass `new_version` from line 119 here instead of re-calling `_get_new_version_number`?

nickclyde (commented):
Thanks Josh! Looks good, just a few notes. Not totally necessary, but a unit test in `test_eicr_augmenter.py` that asserts the new append when absent behavior could be a good addition to the e2e test coverage.

packages/augmentation/src/augmentation/services/eicr_augmenter.py
```diff
@@ -129,43 +147,45 @@ def _handle_related_document_header(self) -> None:
...
-        if not results:
+        result = self._get_element_by_xpath(element, xpath)
+        if result is None:
             raise ValueError(f"Unable to find tag in eICR document for XPath: {xpath}")
-        return results[0]
+        return result
+
+    def _get_element_by_xpath(self, element: Element, xpath: str) -> Element | None:
+        """Get the first matching child element by XPath, or `None` if not found."""
+        results = element.xpath(cda_xpath(xpath), namespaces=CDA_NSMAP)
```
m-goggins, line 180: For my own understanding, are there any scenarios where getting only the first matching child element would be a problem? For example, if there are multiple observations and we're trying to augment the 2nd one, would this be a problem? I thought the xpaths would be specific enough to point to the element we care about so I'm confused about why we're selecting the first element.

m-goggins (approved):
Looks good to me! One question for my own understanding.

packages/augmentation/src/augmentation/services/eicr_augmenter.py
BradySkylight: I think that is how the functionality is for Xpath and the ordering can occur later (or does occur later).  I'll let Josh speak to it, but I think so.

packages/augmentation/src/augmentation/services/eicr_augmenter.py
```diff
@@ -129,43 +147,45 @@ def _handle_related_document_header(self) -> None:
...
+        parent_version_number = self._get_element_by_xpath(
+            self._original_element, "/ClinicalDocument/versionNumber"
+        )
+        if parent_version_number is not None:
+            parent_doc.append(parent_version_number)
+            self._add_previous_element_comment(
+                "input-document-versionNumber ", parent_version_number
+            )
+
+    def _get_required_element_by_xpath(self, element: Element, xpath: str) -> Element:
```
BradySkylight, line 168: What is the difference between the _get_required_element_by_xpath and _get_element_by_xpath?

packages/augmentation/src/augmentation/services/eicr_augmenter.py
```diff
@@ -89,26 +91,42 @@ def augment(self) -> AugmentResult:
...
+        if old_version is not None:
+            self._add_previous_element_comment("new-document-versionNumber ", old_version)
+            self._replace_element("/ClinicalDocument/versionNumber", self._get_new_version_number())
+        else:
+            self._augmented_element.append(new_version)
+            self._add_previous_element_comment("new-document-versionNumber ", new_version)
 
-        new_id = self._get_augmented_tag_by_xpath("/ClinicalDocument/id")
+        new_id = self._get_required_element_by_xpath(
+            self._augmented_element, "/ClinicalDocument/id"
```
BradySkylight, line 130: Isn't the new id and new setId all suppose to be deterministic?  Is this happening somewhere else?  This seems to be pulling the old docId from the eicr and reusing, unless I missed something.

packages/augmentation/src/augmentation/services/eicr_augmenter.py
```diff
@@ -129,43 +147,45 @@ def _handle_related_document_header(self) -> None:
...
+        results = element.xpath(cda_xpath(xpath), namespaces=CDA_NSMAP)
+        if results:
+            return results[0]
+        return None
 
     def _get_old_document_id(self) -> Element:
         """Extract the parent document ID from original eICR document."""
-        parent_doc_id = self._get_original_by_xpath("/ClinicalDocument/id")
+        parent_doc_id = self._get_required_element_by_xpath(
+            self._original_element, "/ClinicalDocument/id"
```
BradySkylight, line 188: Do we already have this in another variable that can be used instead of making the call to get the id again?

packages/augmentation/src/augmentation/services/eicr_augmenter.py
JNygaard-Skylight: It does indeed at it to the end of the `ClinicalDocument`. I do not think there is any strict requirement regarding the ordering, but it seems easy enough to keep the order.

packages/augmentation/src/augmentation/services/eicr_augmenter.py
JNygaard-Skylight: `Element.xpath` returns a list because there technically could be multiple results for any arbitrary XPath. However I think it is safe to say we will always be using this with an XPath where there will only be 1 result or nothing. The XPath should always be specific enough to never return more than 1. But that is a good point and it is worth making that clear in the docs.

packages/augmentation/src/augmentation/services/eicr_augmenter.py
JNygaard-Skylight: `_get_required_element_by_xpath` will raise an error if the element is not found. `_get_element_by_xpath` will return `None` if an element is not found.

packages/augmentation/src/augmentation/services/eicr_augmenter.py
JNygaard-Skylight: This is getting the ID from the augmented eICR, which was updated to the new deterministic ID on line 99.

packages/augmentation/src/augmentation/services/eicr_augmenter.py
JNygaard-Skylight: I'm not sure what you are suggesting. This is the function to get the old document ID.

BradySkylight (approved):
Nice work!  Thanks for answering my questions!

PR #660, opened 2026-06-23
Brady/591 snoinc update to aws

### LINKED ISSUE #591: Modify SNOINC Update Process to run in AWS via GitHub workflow

BradySkylight:
## Description

This PR is fairly large -- I apologize up-front!

The changes in this PR are related to creating an easy way to perform terminology updates, only LOINC Lab Names at this stage though, within an AWS environment via a GitHub Workflow (CI).  The following items were changed:

- New terminology_maintenance.py script added to manage the terminology update process and communication between AWS components and the LOINC API
- A new GitHub Workflow yml script to provide the permissions and access to secrets required to execute the terminology_maintenance.py script
- Updates to loinc.py and general.py to maintain the 'local' versions of functionality that work with files in the local repo as well as abstract other functions to be able to be leverage by both local and cloud-based (AWS) functionality for getting/writing files.
- Updates to existing tests, removal of some out of date tests, and adding a few more tests
- Removal of some old testing artifacts that weren't being leveraged
- Removal of old/former LOINC Update process that was running only with local repo files and print statements (maintenance.py)
- Some terraform variables for a new 'folder' in our AWS S3 Bucket where all extracts, deltas and JSONL files can be stored 'terminologies/'

## Related Issues

Closes #591 

## Additional Notes
Happy to discuss or answer any questions.  Performing test runs with the workflow currently

## Checklist for Reviewers


JNygaard-Skylight (commented):
We're excluding `packages/data-curation/src/data_curation/data_emulation`, because, if I remember correctly, it had a lot of code that we weren't using. Do we need to enable it for this directory now?

BradySkylight:
> We're excluding `packages/data-curation/src/data_curation/data_emulation`, because, if I remember correctly, it had a lot of code that we weren't using. Do we need to enable it for this directory now?

Yes, I think maybe we should so we can leverage this code for the maintenance workflow?  I'm assuming that this is what the following error is related to?  
image

BradySkylight:
> > We're excluding `packages/data-curation/src/data_curation/data_emulation`, because, if I remember correctly, it had a lot of code that we weren't using. Do we need to enable it for this directory now?
> 
> Yes, I think maybe we should so we can leverage this code for the maintenance workflow?

Though, I'm open to arguments otherwise.

JNygaard-Skylight:
> > > We're excluding `packages/data-curation/src/data_curation/data_emulation`, because, if I remember correctly, it had a lot of code that we weren't using. Do we need to enable it for this directory now?
> > 
> > 
> > Yes, I think maybe we should so we can leverage this code for the maintenance workflow?
> 
> Though, I'm open to arguments otherwise.

I mean removing the `extend-exclude` from `packages/data-curation/pyproject.toml`. However, removing that finds 330 errors, 161 can be fixed automatically safely, another 25 can be fixed with `--unsafe-fixes` (I think they're fine, but I have not checked too closely). That still leaves 144 issues, which is a pretty big lift and should probably be left to a seperate ticket.

packages/data-curation/src/data_curation/terminologies/loinc.py
```diff
@@ -41,38 +38,77 @@
...
-EmbeddingRecord = dict[str, object]
 
+def set_loinc_response(
+    terminology_set: str,
+    result: str,
+    message: str,
+    change_log: dict | None = None,
+    embedding_records: list[dict] | None = None,
+) -> TerminologyUpdateResponse:
+    """Defines dictionary for a LOINC Terminology Update Response based upon result and message inputs."""
```
robertandremitchell: should include :param:

.github/scripts/terminology_maintenance.py
```diff
@@ -0,0 +1,344 @@
...
+REGION = AWS_REGION = os.getenv("AWS_REGION", "us-east-1")
+S3_BUCKET = os.getenv("S3_BUCKET", "dibbs-text-to-code")
+TERMINOLOGY_EXTRACT_PREFIX = os.getenv("TERMINOLOGY_EXTRACT_PREFIX", "Terminologies/")
+INGESTION_PREFIX = os.getenv("INGESTION_PREFIX", "ingestion/")
+
+logger = logging.getLogger(__name__)
+_cached_s3_client: BaseClient | None = None
+
+
+def create_s3_client() -> BaseClient:
```
robertandremitchell, line 35: not sure if we want to leverage `lambda-handler` code for this? I know the initial thought was to have all the aws/opensearch-style setup code in one place, but I'm agnostic.

robertandremitchell (commented):
broadly makes sense, a few small comments

BradySkylight:
> > > > We're excluding `packages/data-curation/src/data_curation/data_emulation`, because, if I remember correctly, it had a lot of code that we weren't using. Do we need to enable it for this directory now?
> > > 
> > > 
> > > Yes, I think maybe we should so we can leverage this code for the maintenance workflow?
> > 
> > 
> > Though, I'm open to arguments otherwise.
> 
> I mean removing the `extend-exclude` from `packages/data-curation/pyproject.toml`. However, removing that finds 330 errors, 161 can be fixed automatically safely, another 25 can be fixed with `--unsafe-fixes` (I think they're fine, but I have not checked too closely). That still leaves 144 issues, which is a pretty big lift and should probably be left to a seperate ticket.

I don't think we have to include this code, especially if it's going to cause all the errors.  I was able to get the workflow working without including that package.  So, I think we can skip this.  My assumption was wrong about needed to include it to get the update working.

.github/scripts/terminology_maintenance.py
BradySkylight: It's a good point.  I'm ok either way.  I know this is working now.  Do we want to make another ticket to modify this to use existing code?  Some of it is specific to some of the terminologies files though.  Thoughts?

packages/data-curation/src/data_curation/terminologies/loinc.py
robertandremitchell, line 1: looks like this file has a few coverage gaps, now:
```
packages/data-curation/src/data_curation/terminologies/loinc.py                         230     19    92%   64, 66, 170, 200, 494-525
```

fine tackling in another PR, just flagging to add a ticket

.github/scripts/terminology_maintenance.py
```diff
@@ -0,0 +1,305 @@
...
+    :returns: Terminology Update Response object that contains terminologies, result, and any messages
+    """
+    # get the latest version number and version date of LOINC
+    loinc_version, loinc_version_date = get_loinc_current_version_data()
+    # find the existing TTC LOINC LabNames file to use for comparison
+    current_loinc_file = get_latest_extract_file_name(LAB_NAMES)
+    if current_loinc_file is None:
+        results = load_initial_extract_files()
+        logger.info(results)
+        current_loinc_file = "loinc_lab_names_20260223.csv"
```
robertandremitchell: I think this should be defined near the top of file, or dynamically determined by getting the newest date, if this is a consistent naming convention

.github/scripts/terminology_maintenance.py
JNygaard-Skylight: I think it is reasonable to leave this as a new ticket for look into if we can share the S3 client code.

JNygaard-Skylight (approved):
Approved!

.github/scripts/terminology_maintenance.py
robertandremitchell: I'm good with just adding a backlog ticket. I don't think it matters too much, my main concern is just having more places to manage/maintain AWS if there are breaking changes down the line. probably a low priority given our other work

robertandremitchell (approved):
a few small notes, but mostly makes sense to me

.github/scripts/terminology_maintenance.py
BradySkylight: This is the default/first loinc names extract file that needs to be added to your S3 bucket (terminologies) if no loinc_lab_names_<>.csv file is found.  I will define at the top and use the defined variable instead.

packages/data-curation/src/data_curation/terminologies/loinc.py
BradySkylight: ticket added:  https://app.zenhub.com/workspaces/dibbs-text-to-code-68756bf8ea06bb00319391c0/issues/gh/cdcgov/dibbs-text-to-code/729

.github/scripts/terminology_maintenance.py
BradySkylight: Ticket created:  https://app.zenhub.com/workspaces/dibbs-text-to-code-68756bf8ea06bb00319391c0/issues/gh/cdcgov/dibbs-text-to-code/731

PR #667, opened 2026-06-24
refactor: Remove `ApplicationCode`, simplify augmention process, and a few other little refactors in `eicr_augmenter.py`

JNygaard-Skylight:
## Description
Various Augmenter refactorings:
- Removed `ApplicationCode`, and made the codes value and display module-level constants.
- Removed functions `_get_application_code_value`, and `_get_application_code_display` from `EICRAugmenter`. The constants can be used directly.
- Moved `_cda_element` from `eicr_augmenter.py` to `eicr_utils.py`.
- Moved `document_payload_not_none` from `eicr_augmenter.py` to `eicr_utils.py` and renamed it `parse_document`
- Removed the `augmented_eicr` property from `EICRAugmenter` and `augment()` now returns both in the augmented XML and the metadata in a `AugmentResult` dataclass.
- Created `_replace_element` in `EICRAugmenter`  to handle the replacing element logic, including handling the element tail.

### DIFF
--- a/packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
+++ b/packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py
@@ -310,12 +310,12 @@ def _build_augmented_eicr_output(
     :param augmenter: The initialized EICRAugmenter for the original eICR.
     :return: The augmentation-stage output containing the augmented eICR.
     """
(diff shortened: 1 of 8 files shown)


BradySkylight (approved):
This is a great simplification!  Nice

PR #668, opened 2026-06-24
refactor: use eicr identifiers for original eicr ids

### LINKED ISSUE #629: Use eICR Identifier object for eicr IDs
## Summary
The augmentation metadata and augmentation input models use a string for the eICR IDs.
```
# packages/shared-models/src/shared_models/__init__.py
class TTCAugmenterInput(FrozenBaseModel):

    persistence_id: str
    original_eicr_id: str | None = None
    nonstandard_codes: list[NonstandardCodeInstance] = []
    passthrough: bool = False
    passthrough_reason: PassthroughReason | None = None
```
```
# packages/augmentation/src/augmentation/models/application.py
class Metadata(FrozenBaseModel):
    """Model to hold augmentation metadata."""

    original_eicr_id: str
    augmented_eicr_id: str
    nonstandard_codes: list[NonstandardCodeInstanceMetadata]
    """List of the nonstandard codes TTC attempted to resolve."""
    error: str | None = None
    passthrough: bool = False
    passthrough_reason: PassthroughReason | None = None
```
While using just the `root` attribute of the `id` element is sufficient if the `id` does not have an `extension` attribute, but it is possible for 2 eICRs to have an ID with the same root, but different extensions. Therefore we should use the `CdaInstanceIdentifier` model used by the TTC `Metadata` so that we can record both the root and extension.

robertandremitchell:
## Description
Replaces stringed eicr ids with the CDA II in our workflow.

Biggest change is how empty EICR IDs are handled in terms of erroring out. I'm not sure if this would be considered a breaking change, so happy to rename this PR. I'm not sure how often we will expect that pathing to be exercised of an empty/malformed EICR ID, so I think I lean toward thinking that should be an error, but maybe it's pretty common?

## Related Issues

Closes #629 

## Additional Notes

## Checklist

## Checklist for Reviewers

### DIFF
--- a/e2e/snapshots/test_e2e/test_upload_and_process/patient_alliance/patient_alliance_augmentation_metadata.json
+++ b/e2e/snapshots/test_e2e/test_upload_and_process/patient_alliance/patient_alliance_augmentation_metadata.json
@@ -1,7 +1,19 @@
 {
-  "augmented_eicr_id": "1.2.840.113619.21.1.139060385287897942.5.3^1899729054011780",
+  "augmented_eicr_id": {
+    "assigning_authority_name": null,
+    "displayable": null,
+    "extension": "1899729054011780",
+    "null_flavor": null,
+    "root": "1.2.840.113619.21.1.139060385287897942.5.3"
+  },
   "error": "Unable to find tag in eICR document for XPath: /ClinicalDocument/setId",
   "nonstandard_codes": [],
-  "original_eicr_id": "1.2.840.113619.21.1.139060385287897942.5.3^1899729054011780",
+  "original_eicr_id": {
+    "assigning_authority_name": null,
+    "displayable": null,
+    "extension": "1899729054011780",
+    "null_flavor": null,
+    "root": "1.2.840.113619.21.1.139060385287897942.5.3"
+  },
   "passthrough_reason": "augmentation_exception"
 }
--- a/e2e/snapshots/test_e2e/test_upload_and_process/patient_alliance/patient_alliance_augmentation_metadata.json
+++ b/e2e/snapshots/test_e2e/test_upload_and_process/patient_alliance/patient_alliance_augmentation_metadata.json
@@ -7,7 +7,6 @@
     "root": "1.2.840.113619.21.1.139060385287897942.5.3"
   },
   "error": "Unable to find tag in eICR document for XPath: /ClinicalDocument/setId",
-  "nonstandard_codes": [],
   "original_eicr_id": {
     "assigning_authority_name": null,
     "displayable": null,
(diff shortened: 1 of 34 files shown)


JNygaard-Skylight:
To try to address what to do when the ID is empty:

From the [eICR IG](https://www.hl7.org/implement/standards/product_brief.cfm?product_id=436). All eICRs should have an ID. If it does not it is an invalid eICR so I think it is reasonable to throw an error as soon as possible. In addition, the eICR spec. says:
> This id SHALL be a globally unique identifier for the document (CONF:1198-9991).

It does not specify if a root or extension is required (neither are required for an ID according to the [CDA IG](https://hl7.org/cda/stds/core/StructureDefinition-II.html)), but it reasonable to assume that an ID cannot be a globally unique identifier without a root or an extension or both. Therefore I think again it is reasonable to throw an error in the unlikely case we see an empty ID.

While I do see the argument for doing our best to add standardized codes despite whatever issues an eICR may have, my current position is that we should not be building to accommodate eICRs outside of the specification.

BradySkylight:
I agree with Josh in regard to an eICR that doesn't have an ID.  If one is not defined, that is a malformed message and I would doubt that it may even reach TTC.  I would think that it would fail schematron validation in a way that would prevent it from continuing down the pipeline.

robertandremitchell:
that makes sense to me, thanks @BradySkylight and @JNygaard-Skylight !

e2e/test_e2e.py
```diff
@@ -315,6 +315,37 @@ def _assert_augmented_eicr_retains_original_content(
     )
 
 
+def _read_ttc_metadata_snapshot(eicr_id: str) -> object:
```
JNygaard-Skylight: This looks good, but could it be pulled out into its own PR? If you think that would be too much work for not enough benefit, I am fine with that.

packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
```diff
@@ -108,14 +114,15 @@ def test_handler_success(
...
-        assert metadata["original_eicr_id"] == EXPECTED_ORIGINAL_EICR_ID
-        assert metadata["augmented_eicr_id"] == EXPECTED_AUGMENTED_EICR_ID
-        assert metadata["augmented_eicr_id"] != metadata["original_eicr_id"]
-        assert metadata["augmented_eicr_id"] != TEST_PERSISTENCE_ID
-        assert metadata["passthrough_reason"] is None
+        metadata = Metadata.model_validate_json(metadata_raw)
+        assert metadata.original_eicr_id == EXPECTED_ORIGINAL_EICR_ID
+        assert metadata.augmented_eicr_id == EXPECTED_AUGMENTED_EICR_ID
+        assert metadata.augmented_eicr_id != metadata.original_eicr_id
+        assert metadata.augmented_eicr_id.root != TEST_PERSISTENCE_ID
```
JNygaard-Skylight: What is the purpose of asserting that the root and extension do not equal the persistence ID?

e2e/test_e2e.py
robertandremitchell: no, I think that makes sense to do since this is a separate concern. Should have time to split this off this morning.

packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
robertandremitchell: I think this was a leftover from the test rewrite. definitely not needed anymore.

e2e/snapshots/test_e2e/test_upload_and_process/patient_alliance/patient_alliance_augmentation_metadata.json
```diff
@@ -1,92 +1,19 @@
...
-  "original_eicr_id": "1.2.840.113619.21.1.139060385287897942.5.3",
-  "passthrough_reason": null
+  "augmented_eicr_id": {
+    "assigning_authority_name": null,
+    "displayable": null,
+    "extension": "1899729054011780",
+    "null_flavor": null,
+    "root": "1.2.840.113619.21.1.139060385287897942.5.3"
+  },
+  "error": "Unable to find tag in eICR document for XPath: /ClinicalDocument/setId",
```
JNygaard-Skylight: It looks like the handling of missing `setId` has been broken

robertandremitchell: good catch, should be resolved now!

packages/augmentation/src/augmentation/services/eicr_augmenter.py
```diff
@@ -165,15 +174,24 @@ def _handle_related_document_header(self) -> None:
...
 
-        While it is possible that an XPath could return multiple results, we are currently only using this to look for elements that either should only have at most one instance (such as `effectiveTime`, `setId`), or the XPath should be specific enough to uniquely identify an element (such as the XPath to an `observation`).
-        """
-        result = self._get_element_by_xpath(element, xpath)
-        if result is None:
-            raise ValueError(f"Unable to find tag in eICR document for XPath: {xpath}")
-        return result
+    def _get_augmented_tag_by_xpath(self, xpath: str) -> Element:
+        """Get element from the augmented eICR by XPath."""
+        return self._get_required_element_by_xpath(self._augmented_element, xpath)
```
BradySkylight, line 183: Were these functions rename or removed during a previous PR by Josh?

robertandremitchell: I can take a second look to confirm nothing got lost with the merge of Josh's PR, it got a bit complex to disentangle because I think we made a lot of a similar edits.

robertandremitchell: yeah, the track changes just got messed up here. Code exists here, unchanged: https://github.com/CDCgov/dibbs-text-to-code/pull/668/changes#diff-6def2d44248a0a34fa6ba9562c57393b269135063fd40b7a0b35ad824354c4dcR228-R237

(lines 228-237)

JNygaard-Skylight (approved):
Approved!

PR #672, opened 2026-06-25
Release v0.2.5: fix Hugging Face Hub lookup crash on baked-in local model paths

nickclyde:
## Description

Hotfix for the production outage reported by APHL on 0.2.4: every eICR submitted to the TTC Lambda failed and went to the DLQ.

### Root cause

`get_model_info()` (added in #627) calls `huggingface_hub.model_info()` at **module import time** in `embedder.py` and `reranker.py`. In production the retriever/reranker models are baked into the container image and referenced by a **local filesystem path** (`RETRIEVER_MODEL_PATH=/opt/retriever_model`), which is not a valid Hub repo id. `model_info()` raised:

```
HFValidationError: Repo id must be in the form 'repo_name' or 'namespace/repo_name': '/opt/retriever_model'.
```

This crashed the Lambda during import — before any message could be processed — so all messages went to the DLQ.

Two compounding problems:
1. A local path is not a valid Hub repo id, so validation fails.
2. The lookup hits the Hub over the internet, but the production VPC has **no internet route** — so the call could never succeed in prod regardless of the value.

The `Loading weights: 100%` line in the prod logs confirms the model itself loaded fine from disk; only the metadata lookup crashed.

### Fix

`get_model_info()` now:
- Skips the Hub lookup for local directory paths and returns just the `id` (no network call).
- Degrades gracefully on `HFValidationError` instead of crashing import.
- Still raises on `RepositoryNotFoundError` (unchanged behavior).

`ModelInfo` fields are all `Optional` and the Lambda already treats `model_info` as optional, so metadata simply omits the Hub timestamps for baked-in models (which it could not fetch in the prod VPC anyway).

### Why CI didn't catch it

In tests, `RETRIEVER_MODEL_PATH` is unset, so `TTC_RETRIEVER` falls back to the `NCHS/ttc-retriever-mvp` repo id — a valid id — and the lookup succeeds. The failure only manifests with a local model path, i.e. exclusively in the deployed container.

## Changes

- `get_model_info()`: short-circuit local paths; catch `HFValidationError`.
- Two regression tests: local-path skips the Hub lookup; invalid repo id degrades instead of raising.
- Version bump `0.2.4` → `0.2.5` (root `pyproject.toml` + `uv.lock`) to cut the hotfix release.

## Testing

- `packages/text-to-code` unit suite: 81 passed.
- `just ruff`: clean. `just ty`: no new diagnostics (2 pre-existing `fastapi` errors in `local_server.py` are unrelated).
- Direct repro: `get_model_info('/opt')` now returns degraded info instead of raising.

### DIFF
--- a/packages/text-to-code/src/text_to_code/services/utils.py
+++ b/packages/text-to-code/src/text_to_code/services/utils.py
@@ -1,9 +1,14 @@
+import logging
+import os
+
 from huggingface_hub import errors, model_info
 
 from shared_models import DataField
 from text_to_code.models.labs import BaseLabField
 from text_to_code.models.model_info import ModelInfo
 from text_to_code.models.registry import EICR_REGISTRY
+
+logger = logging.getLogger(__name__)
 
 ConfigType = type[BaseLabField]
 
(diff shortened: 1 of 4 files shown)


bamader (approved):
Good catch and fix!

robertandremitchell (approved):
makes sense to me

PR #673, opened 2026-06-26
Changing exception logic for eicr/schematron processing

### LINKED ISSUE #670: Replace broad `except Exception` that can silently drop eICR data
## Problem

Two core processors wrap large blocks in bare `except Exception` (both *do* call `logger.exception`, so tracebacks are logged — but the granularity hides real failures):

1. [`packages/text-to-code/.../eicr_processor.py:47-76`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/packages/text-to-code/src/text_to_code/services/eicr_processor.py#L47-L76) — the `try` wraps the **entire** nested node/sub-xpath loop. A single malformed sub-node aborts extraction for the whole `(base_xpath, data_field)` and returns whatever was collected so far. Early returns here could be masking systematic extraction failures.
2. [`packages/text-to-code/.../schematron_processor.py:74-123`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/packages/text-to-code/src/text_to_code/services/schematron_processor.py#L74-L123) — per-`validationResult` `try/except … continue`. [Line 92](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/packages/text-to-code/src/text_to_code/services/schematron_processor.py#L92) (`assertion_id_elem.text.strip()`) assumes `assertionID` exists; when it doesn't, the resulting `AttributeError` is swallowed and the error is silently dropped.

## Proposed scope

- Narrow the `try` in `get_text_candidates` to per-sub-node so one bad node doesn't kill the batch; catch specific exceptions (`etree.XPathError`, `AttributeError`, `ValueError`).
- In `schematron_processor`, guard `assertion_id_elem is None` explicitly (like the other `*_elem` None checks at [`:83-89`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/packages/text-to-code/src/text_to_code/services/schematron_processor.py#L83-L89)) instead of relying on the broad except; narrow the catch.
- Add a metric/structured-log counter distinguishing "no candidates (expected)" from "extraction raised" so the passthrough investigation has signal.

## Acceptance criteria

- Tests cover: malformed sub-node mid-loop (other candidates still returned); a `validationResult` missing `assertionID` (handled deterministically, logged, not crashed).
- No bare `except Exception` remains in these two functions.

robertandremitchell:
## Description

This is to get more info/data when Exceptions occur.

## Related Issues

Closes #670 

## Additional Notes

## Checklist

## Checklist for Reviewers

### DIFF
--- a/packages/text-to-code/src/text_to_code/models/eicr.py
+++ b/packages/text-to-code/src/text_to_code/models/eicr.py
@@ -35,4 +35,4 @@ class TextCandidateExtractionLogContext:
 
     base_xpath: str
     data_field: DataField
-    sub_xpaths: list[str]
+    sub_xpaths: list[LabXPaths]
--- a/packages/text-to-code/src/text_to_code/models/eicr.py
+++ b/packages/text-to-code/src/text_to_code/models/eicr.py
@@ -1,6 +1,7 @@
+from dataclasses import dataclass
 from enum import StrEnum
 
-from shared_models import CdaInstanceIdentifier, FrozenBaseModel
+from shared_models import CdaInstanceIdentifier, DataField, FrozenBaseModel
 
 
 class LabXPaths(StrEnum):
@@ -26,3 +27,12 @@ class Metadata(FrozenBaseModel):
 
     eicr_id: CdaInstanceIdentifier | None
     eicr_vendor: str | None = None
+
+
+@dataclass(frozen=True)
+class TextCandidateExtractionLogContext:
+    """Context for logging text candidate extraction errors and summaries."""
+
+    base_xpath: str
+    data_field: DataField
+    sub_xpaths: list[str]
--- a/packages/text-to-code/src/text_to_code/services/eicr_processor.py
+++ b/packages/text-to-code/src/text_to_code/services/eicr_processor.py
@@ -87,7 +87,7 @@ class EicrProcessor:
                             sub_node=sub_node,
                             xpath=sub_xpath,
                         )
-                    except (etree.XPathError, AttributeError, ValueError):
+                    except (etree.XPathError, AttributeError, SyntaxError, ValueError):
                         extraction_error_count += 1
                         self._log_text_candidate_extraction_error(
                             log_context=log_context,
--- a/packages/text-to-code/src/text_to_code/services/eicr_processor.py
+++ b/packages/text-to-code/src/text_to_code/services/eicr_processor.py
@@ -5,7 +5,7 @@ from lxml.etree import Element
 
 from shared_models import CdaInstanceIdentifier, DataField
 from text_to_code.models import Candidate
-from text_to_code.models.eicr import Metadata
+from text_to_code.models.eicr import Metadata, TextCandidateExtractionLogContext
 from text_to_code.services.utils import get_config_for_data_field
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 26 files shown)


JNygaard-Skylight (approved):
Let's good to me!

packages/text-to-code/src/text_to_code/models/eicr.py
```diff
@@ -26,3 +27,12 @@ class Metadata(FrozenBaseModel):
...
     eicr_vendor: str | None = None
+
+
+@dataclass(frozen=True)
+class TextCandidateExtractionLogContext:
+    """Context for logging text candidate extraction errors and summaries."""
+
+    base_xpath: str
+    data_field: DataField
+    sub_xpaths: list[LabXPaths]
```
BradySkylight, line 38: This is a good idea to have to understand where a failure may have occurred.  Nice!

packages/text-to-code/src/text_to_code/services/eicr_processor.py
```diff
@@ -34,6 +34,8 @@ def get_text_candidates(self, base_xpath: str, data_field: DataField) -> list[Ca
             the specified data field for TTC processing.
         """
         candidates: list[Candidate] = []
+        extraction_error_count = 0
+        no_candidate_count = 0
         # first get data field config settings - this acts
         # as a validation of correct data field being passed
         config_settings = get_config_for_data_field(data_field)
```
BradySkylight, line 41: Did we, or are we going to, change this to not require trying to get information from a config?  Just keeping it since it's working I assume?

BradySkylight (approved):
This looks good to me.  I had a couple of clarifying questions I think, but nothing blocking.

packages/text-to-code/src/text_to_code/services/eicr_processor.py
robertandremitchell: yeah, I think it makes sense to keep for now, but could be worth revisiting. curious @nickclyde 's thoughts if this was intended to be part of this ticket?

nickclyde: From what I can tell, this function just checks whether it's a resulted or ordered field, and adds some extra config based on which? Is there a risk to keeping this line?

packages/text-to-code/src/text_to_code/services/eicr_processor.py
```diff
@@ -43,37 +45,160 @@ def get_text_candidates(self, base_xpath: str, data_field: DataField) -> list[Ca
...
+                    continue
+
+                for sub_node in sub_nodes:
+                    try:
+                        candidate_added = self._append_text_candidates_from_sub_node(
+                            candidates=candidates,
+                            sub_node=sub_node,
+                            xpath=sub_xpath,
+                        )
+                    except (etree.XPathError, AttributeError, ValueError):
```
nickclyde: Can we add `SyntaxError` here as well? `resolve_reference` runs `self._xml_root.find('.//*[@ID="..."]')`. If a `<reference>`'s resolved value contains a `"` (for example a `value` of `#foo"`), lxml's `.find()` raises `SyntaxError: invalid predicate`, so we should be sure to catch that as well.

nickclyde (commented):
Thanks Rob!! This looks great, just one suggestion.

packages/text-to-code/src/text_to_code/services/eicr_processor.py
robertandremitchell: updated!

Issue #655, opened 2026-06-23
just command to force build validation files.


JNygaard-Skylight:

JNygaard-Skylight:
This was resolved in PR #675

PR #675, opened 2026-06-30
fix: Adding code to regenerate validation test files

### LINKED ISSUE #638: Determine if temp validation files need to be regenerated
There is a relatively minor issue where because the temp files generated in order to run `validation/` are cached that can cause local tests to fail. 

This ticket would either
1. Generate new temp files with each test suite run to avoid this mismatch
2. Use a checker of some kind to determine if new temp files are needed

Maybe there's another solution. Not a very high pressure ticket though, given that it is easily resolved by just deleting and rerunning validation.

### Acceptance Criteria

**Scenario 1: Automated detection of stale validation files**
- **Given** the validation test suite is initiated
- **When** the system detects that the existing temporary validation files are out of sync with the current source code or configuration
- **Then** the system should automatically regenerate the necessary temporary files before executing the tests

**Scenario 2: Successful test execution without manual intervention**
- **Given** a local environment with cached temporary validation files that would previously cause a mismatch failure
- **When** the user runs the `validation/` test suite
- **Then** the tests should pass successfully without requiring the user to manually delete the temporary files

**Scenario 3: Optimization of test suite performance (Optional/Conditional)**
- **Given** the temporary validation files are already up-to-date and match the current source
- **When** the validation test suite is executed
- **Then** the system should skip the regeneration step to minimize test execution time (if the "checker" approach is implemented)

robertandremitchell:
## Description

This adds code to `validation/main.py` that checks the last modified date of the two dependent files to see if new temps are needed. This should be testable locally by making a nonsense change (adding a `<>` to the blank line 3 of the `packages/validation/src/validation/eicr-validator/schematron/APHL_TextToCodeSchematron_09252025.sch`) and then confirming that new temp files were generated in `packages/validation/src/validation/eicr-validator/output`.

This also adds a just command to manually remove the files to generate new ones, regardless of whether the files had changed.

These changes are needed to address issues that can arise where locally, the temp files become stale as a result of changes to the schematron or xslt compile file.

## Related Issues

Closes #638, #655 

## Additional Notes

## Checklist

## Checklist for Reviewers

### DIFF
--- a/.justscripts/just/test.just
+++ b/.justscripts/just/test.just
@@ -15,3 +15,8 @@ all *ARGS:
 [doc("Run tests with coverage, specifying the package to measure coverage for")]
 coverage *ARGS:
     uv run pytest {{ ARGS }} --cov={{ARGS}} --cov-report=term-missing
+
+[doc("Regenerate validation temp files and run validation tests")]
+validation:
+    uv run python -c "from pathlib import Path; output = Path('validation/src/validation/eicr-validator/output'); [file.unlink(missing_ok=True) for file in (output / 'stage1.sch.tmp', output / 'stage2.sch.tmp', output / 'validator.xsl.tmp', output / 'voc_ttc.xml')]"
+    uv run pytest packages/validation
--- a/.justscripts/just/test.just
+++ b/.justscripts/just/test.just
@@ -18,5 +18,5 @@ coverage *ARGS:
 
 [doc("Regenerate validation temp files and run validation tests")]
 validation:
-    uv run python -c "from pathlib import Path; output = Path('validation/src/validation/eicr-validator/output'); [file.unlink(missing_ok=True) for file in (output / 'stage1.sch.tmp', output / 'stage2.sch.tmp', output / 'validator.xsl.tmp', output / 'voc_ttc.xml')]"
+    uv run python -c "from pathlib import Path; output = Path('packages/validation/src/validation/eicr-validator/output'); [file.unlink(missing_ok=True) for file in (output / 'stage1.sch.tmp', output / 'stage2.sch.tmp', output / 'validator.xsl.tmp', output / 'voc_ttc.xml')]"
     uv run pytest packages/validation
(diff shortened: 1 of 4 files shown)


.justscripts/just/test.just
```diff
@@ -14,4 +14,9 @@ all *ARGS:
 
 [doc("Run tests with coverage, specifying the package to measure coverage for")]
 coverage *ARGS:
-    uv run pytest {{ ARGS }} --cov={{ARGS}} --cov-report=term-missing
\ No newline at end of file
+    uv run pytest {{ ARGS }} --cov={{ARGS}} --cov-report=term-missing
+
+[doc("Regenerate validation temp files and run validation tests")]
+validation:
+    uv run python -c "from pathlib import Path; output = Path('validation/src/validation/eicr-validator/output'); [file.unlink(missing_ok=True) for file in (output / 'stage1.sch.tmp', output / 'stage2.sch.tmp', output / 'validator.xsl.tmp', output / 'voc_ttc.xml')]"
```
nickclyde: I think this path is missing the `packages/` prefix. `test.just` sets `working-directory := '../..'` (the repo root), and from there the output dir lives at `packages/validation/src/validation/eicr-validator/output`. The `pytest` line right below already uses the `packages/`-prefixed path. Could we change it to `packages/validation/src/validation/eicr-validator/output`?

packages/validation/tests/test_validation.py
```diff
@@ -258,3 +259,69 @@ def test_build_schematron_report_xml_normalizes_real_location():
...
+
+    old_time = 1_000_000_000
+    new_time = 1_000_000_100
+
+    for output_file in (stage1_output, stage2_output, validator_output, voc_output):
+        output_file.touch()
+        output_file.chmod(0o644)
+        output_file.write_text(output_file.read_text())
+        output_file.stat()
+        output_file.touch()
```
nickclyde: I think this setup loop has some redundant steps. The files already have content written above, and the only operation that affects the test is the final `os.utime(...)` that backdates the mtime. Could we simplify the loop body to just the `os.utime(output_file, (old_time, old_time))` call to make the intent clearer?

nickclyde (commented):
Looks good! Just a few notes:

packages/validation/tests/test_validation.py
robertandremitchell: agreed. I think this does a better job of simplifying the denseness of this code. lmk!

.justscripts/just/test.just
robertandremitchell: updated!

nickclyde (approved):
Thanks Rob!!

PR #679, opened 2026-06-30
refactor: Centralize code systems

JNygaard-Skylight:
## Description

Create constants for Loinc and SNOMED identifiers in the shared model package.

## Related Issues

Closes #671 

## Additional Notes

[Add any additional context or notes that reviewers should know about.]

<--------------------- REMOVE THE LINES BELOW BEFORE MERGING --------------------->

## Checklist

Please review and complete the following checklist before submitting your pull request:

- [ ] I have ensured that the pull request is of a manageable size, allowing it to be reviewed within a single session.
- [ ] I have reviewed my changes to ensure they are clear, concise, and well-documented.
- [ ] I have updated the documentation, if applicable.
- [ ] I have added or updated test cases to cover my changes, if applicable.
- [ ] I have minimized the number of reviewers to include only those essential for the review.

## Checklist for Reviewers

Please review and complete the following checklist during the review process:

- [ ] The code follows best practices and conventions.
- [ ] The changes implement the desired functionality or fix the reported issue.
- [ ] The tests cover the new changes and pass successfully.
- [ ] Any potential edge cases or error scenarios have been considered.

PR #683, opened 2026-07-01
refactor: Changing how SQS messages are logged in lambdas

### LINKED ISSUE #669: SQS message loss on per-record failure
## Problem

Both Lambda handlers catch per-record exceptions, append to a `failures` list, and then **return a normal dict** instead of raising:
- [`packages/text-to-code-lambda/.../lambda_function.py:50-89`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py#L50-L89)
- [`packages/augmentation-lambda/.../lambda_function.py:44-79`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py#L44-L79)

With an SQS event source mapping, SQS deletes a message whenever the handler **returns normally** — the return value is ignored unless the mapping sets `function_response_types = ["ReportBatchItemFailures"]` (verified **absent** in [`terraform/main.tf:739`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/terraform/main.tf#L739), [`:832`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/terraform/main.tf#L832)). So every record placed in `failures` is acked and deleted: never retried, never sent to the DLQ (`maxReceiveCount = 3` → DLQ, [`terraform/main.tf:685`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/terraform/main.tf#L685), [`:778`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/terraform/main.tf#L778)). The DLQ only ever fires when the handler *raises* — e.g. OpenSearch client creation at [`text-to-code .../lambda_function.py:43`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py#L43), which is *outside* the try/except. The `failures`/`statusCode` payloads are dead for routing.

This is not a multi-record / partial-batch bug. Both mappings use `batch_size = 1` ([`terraform/main.tf:742`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/terraform/main.tf#L742), [`:835`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/terraform/main.tf#L835)), so every production invocation contains exactly one record, and that single failing record is silently dropped. The multi-record loop and `failures`/`successes` accounting never see >1 record in prod.

This went unnoticed because the lost-message path only triggers when the per-record body raises; happy-path eICRs never exercise it. The TTC Lambda absorbs most exceptions into a `TTC_EXCEPTION` passthrough counted as success ([`lambda_function.py:60-64`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py#L60-L64)); a record reaches `failures` only if the passthrough write also fails. So an eICR file alone cannot demonstrate the drop on TTC. The augmentation Lambda has no such absorption — any raise in `_process_record` gets dropped. The symptom is a *negative* (no DLQ delivery) that you'd only catch by asserting retry/DLQ on a deliberately failing input. Furthermore, the current tests encode the bug as expected behavior: [`test_handler_error_missing_eicr`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/packages/augmentation-lambda/tests/test_augmentation_lambda_function.py#L600) feeds a single-record event, deletes the original eICR so `_load_original_eicr` raises `FileNotFoundError`, and asserts the handler *returns* `statusCode 207` + `num_failure_eicrs: 1` (snapshot [`handler_error_missing_eicr_result.json`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/packages/augmentation-lambda/tests/snapshots/test_augmentation_lambda_function/test_handler_error_missing_eicr/handler_error_missing_eicr_result.json)). No exception, no `batchItemFailures`, so SQS would delete that message.

 ## Proposed scope

- Add `function_response_types = ["ReportBatchItemFailures"]` to both event-source mappings.
- Change both handlers to return `{"batchItemFailures": [{"itemIdentifier": <message_id>}, ...]}` for failed records (the AWS-required shape). Keep structured logging of counts.
- Standardize success/partial status logging (drop the 207-vs-200 inconsistency: [`augmentation .../lambda_function.py:58`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/packages/augmentation-lambda/src/augmentation_lambda/lambda_function.py#L58) vs [`text-to-code .../lambda_function.py:68`](https://github.com/CDCgov/dibbs-text-to-code/blob/a15881114eff9876b34a1929b7d88f98dc64c492/packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py#L68)).
- Update `test_handler_error_missing_eicr` / `test_handler_error_missing_ttc_output` and their result snapshots to assert the failed message appears in `batchItemFailures`.

## Acceptance criteria

- A single-record event whose record fails (and, for TTC, whose passthrough write also fails) is retried per `maxReceiveCount` and lands in the DLQ — not silently deleted.
- In a (future) multi-record batch, only failed records are retried; succeeded ones are acked.
- Tests assert the `batchItemFailures` payload; the updated snapshots reflect it.

robertandremitchell:
## Description

This addresses a change in how SQS messages are logged in lambdas and updates the terraform appropriately to ensure it actually logs when and where failures occur. 

Tooled around a bit Thursday and this morning on this PR, which merged some changes into batchItemFailures: https://github.com/CDCgov/dibbs-text-to-code/pull/685.

Not sure if this would count as a refactor, but given that it would change the shape of how lambda data is potentially returned and processed, it seems like it would.

## Related Issues

Closes #669 

## Additional Notes

## Checklist

## Checklist for Reviewers


packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
```diff
@@ -11,8 +11,6 @@
 
 import lambda_handler
 from shared_models import (
-    LOINC_NAME,
-    LOINC_OID,
```
nickclyde, line 15: These were already unused (the `Code(...)` construction further down hardcodes the OID and name), so removing them is technically correct, but could we instead use `LOINC_OID` and `LOINC_NAME` at that hardcoded spot (around line 447) to stay consistent with #679?

terraform/main.tf
```diff
@@ -737,9 +737,10 @@ resource "aws_cloudwatch_event_target" "augmentation_sqs_target" {
 #############
 
 resource "aws_lambda_event_source_mapping" "augmentation_sqs" {
-  event_source_arn = aws_sqs_queue.augmentation_queue.arn
-  function_name    = aws_lambda_function.augmentation_lambda.arn
-  batch_size       = 1
+  event_source_arn        = aws_sqs_queue.augmentation_queue.arn
+  function_name           = aws_lambda_function.augmentation_lambda.arn
+  batch_size              = 1
+  function_response_types = ["ReportBatchItemFailures"]
```
nickclyde, line 743: With `ReportBatchItemFailures` enabled, failed messages will now actually land in the DLQs after 3 receives, which is great. I do not think we have any CloudWatch alarms on the DLQs yet, though, so those failures could still go unnoticed. Could we file a follow-up issue for an alarm on `ApproximateNumberOfMessagesVisible` for both DLQs?

packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
```diff
@@ -27,6 +27,7 @@
 EXPECTED_AUGMENTED_EICR_ID = CdaInstanceIdentifier(
     root="d44dc1c6-8a0c-5236-906e-12f6475589ec", extension=None
 )
+NO_BATCH_ITEM_FAILURES = {"batchItemFailures": []}
```
nickclyde: I believe `SUCCESS_CODE` (line 21) is unused now that the status code assertions are gone, so we could remove it.

nickclyde (commented):
Awesome work, looks like this should fix #669!

RE: your question about whether `batchItemFailures` needs to be the only thing returned: I think it should be okay to include extra information. From what I can tell, Lambda only inspects the `batchItemFailures` key, and the conditions for treating the batch as a complete failure are an invalid JSON response or a bad `itemIdentifier` (null, empty, wrong key name, or a message ID that does not exist in the batch). Extra top-level fields are not on that list, so I believe the #685 approach of returning more metadata alongside `batchItemFailures` should work fine.

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
robertandremitchell: ahhh, I missed that, just saw ruff kill these imports. Updating!

terraform/main.tf
robertandremitchell: Ticket created! https://app.zenhub.com/workspaces/dibbs-text-to-code-68756bf8ea06bb00319391c0/issues/gh/cdcgov/dibbs-text-to-code/686

packages/augmentation-lambda/tests/test_augmentation_lambda_function.py
robertandremitchell: updated!

packages/text-to-code-lambda/src/text_to_code_lambda/lambda_function.py
robertandremitchell: updated!

nickclyde (approved):
Looks great, thanks Rob!!

PR #691, opened 2026-07-08
fix: Adding CloudWatch alarms for lambdas dlq

### LINKED ISSUE #686: Add CloudWatch Alarms for Lambda DLQ Message Visibility
With ReportBatchItemFailures enabled (https://github.com/CDCgov/dibbs-text-to-code/pull/683), failed SQS messages will now be retried correctly and sent to the relevant DLQs after three receives. However, we do not currently appear to have CloudWatch alarms monitoring those DLQs, so messages could accumulate without anyone being notified. Add CloudWatch alarms on ApproximateNumberOfMessagesVisible for both DLQs. The alarm should notify the appropriate team/channel whenever either DLQ has visible messages.

robertandremitchell:
## Description

This PR adds CloudWatch alarms for the augmentation Lambda and TTC Lambda dead-letter queues (DLQs). These alarms will trigger when either DLQ has visible messages, which should help surface failed messages instead of allowing them to accumulate unnoticed.

CloudWatch alarms:
https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#alarmsV2:?~(alarmListFilter~(tokens~(~(propertyKey~'alarmState~operator~'*3d~value~'OK))~operation~'and))

Right now, the alarms publish to an SNS topic when triggered. I also added an interim SQS subscription so the alarm messages have an AWS-managed destination while we decide what the final notification path should be.

These alerts will be sent to our Slack channel: https://us-east-2.console.aws.amazon.com/chatbot/home?region=us-east-1#/chat-clients/slack/workspaces/T40A7NFB7/configurations

The two new constants `TF_VAR_slack_team_id` and `TF_VAR_slack_channel_id` manage the messages going to our eng slack channel.

I’m also not sure yet how noisy these alarms will be. The current setup gives us a safe place to collect alarm messages while we determine the right audience and notification channel.

## Related Issues

Closes #686 

## Additional Notes

## Checklist

## Checklist for Reviewers

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -676,7 +676,8 @@ resource "aws_sns_topic" "dlq_alarm_notifications" {
 }
 
 resource "aws_sqs_queue" "dlq_alarm_notifications_queue" {
-  name = "ttc-dlq-alarm-notifications-queue"
+  name                      = "ttc-dlq-alarm-notifications-queue"
+  message_retention_seconds = 1209600
 
   tags = local.tags
 }
@@ -713,8 +714,9 @@ resource "aws_sns_topic_subscription" "dlq_alarm_notifications_queue" {
 #############
 
 resource "aws_sqs_queue" "augmentation_dlq" {
-  name = "${var.augmentation_lambda_function_name}-dlq"
-  tags = local.tags
+  name                      = "${var.augmentation_lambda_function_name}-dlq"
+  message_retention_seconds = 1209600
+  tags                      = local.tags
 }
 
 resource "aws_cloudwatch_metric_alarm" "augmentation_dlq_visible_messages" {
@@ -829,8 +831,9 @@ resource "aws_iam_role_policy" "augmentation_sqs_policy" {
 #############
 
 resource "aws_sqs_queue" "ttc_input_dlq" {
-  name = "${var.lambda_function_name}-dlq"
-  tags = local.tags
+  name                      = "${var.lambda_function_name}-dlq"
+  message_retention_seconds = 1209600
+  tags                      = local.tags
 }
 
 resource "aws_cloudwatch_metric_alarm" "ttc_input_dlq_visible_messages" {
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -675,6 +675,39 @@ resource "aws_sns_topic" "dlq_alarm_notifications" {
   tags = local.tags
 }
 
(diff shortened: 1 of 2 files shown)


JNygaard-Skylight (approved):
LTGTM!

terraform/main.tf
```diff
@@ -669,6 +669,45 @@ resource "aws_lambda_function" "augmentation_lambda" {
   tags = { Name = var.augmentation_lambda_function_name }
 }
 
+resource "aws_sns_topic" "dlq_alarm_notifications" {
+  name = "ttc-dlq-alarm-notifications"
+
+  tags = local.tags
+}
+
+resource "aws_sqs_queue" "dlq_alarm_notifications_queue" {
```
nickclyde, line 688: Since nothing consumes this queue yet, alarm messages will sit here until SQS's default 4-day retention silently expires them. Could we bump `message_retention_seconds` to the 14-day max (1209600) so notifications don't quietly disappear over a long weekend?

terraform/main.tf
```diff
@@ -678,6 +717,28 @@ resource "aws_sqs_queue" "augmentation_dlq" {
   tags = local.tags
 }
 
+resource "aws_cloudwatch_metric_alarm" "augmentation_dlq_visible_messages" {
```
nickclyde, line 767: Both DLQs use the default 4-day retention, so if nobody acts within 4 days the failed messages expire, the alarm auto-resolves, and the eICRs are lost. Now that we're alarming on these queues, I think it would pair well to set `message_retention_seconds = 1209600` (14 days) on `augmentation_dlq` and `ttc_input_dlq` to widen the response window. Happy for that to be a follow-up if you'd rather keep this PR alarm-only.

nickclyde (commented):
Nice addition Rob! Since these are operational signals about our own Lambdas failing (not something APHL can act on), I think it makes most sense to send these alarms to our eng channel in Slack. Noise-wise, since the alarm latches in ALARM until the queue is drained, it should be at most one notification per incident rather than one per message. A few thoughts below:

terraform/main.tf
robertandremitchell: updated!

terraform/main.tf
robertandremitchell: updated!

nickclyde (approved):
Awesome work, thanks Rob!!

PR #711, opened 2026-07-20
Handle params in just test coverage

JNygaard-Skylight:
the `just test coverage` command would not correct handle additional parameters, as it is only expecting a path. As an example `just test coverage ./packages/lambda-handler -v` would result in`uv run pytest ./packages/lambda-handler -v --cov=./packages/lambda-handler -v --cov-report=term-missing`.

With this fix:
```
❯ just test coverage ./packages/lambda-handler -v
uv run pytest ./packages/lambda-handler -v --cov=./packages/lambda-handler --cov-report=term-missing
```


robertandremitchell (approved):
good catch!

PR #713, opened 2026-07-21
deps: uv upgrade and pre-commit update

JNygaard-Skylight:
## Description

Ran `uv sync --all-packages --all-groups --upgrade ` and `pre-commit autoupdate`. Hopefully this resolves some of the security issues that just popped up


robertandremitchell (approved):
local trivy scan looks good to me

PR #714, opened 2026-07-21
Deploy the demo frontend and synchronous API to AWS

nickclyde:
## Summary

#619 added the interactive demo (static page + synchronous `/text-to-code` API) with local-only tooling and noted Terraform would follow. This PR deploys it: a single CloudFront distribution behind Basic auth serves the frontend from a private S3 bucket and routes `POST /text-to-code` to a new `ttc-api-lambda`.

```
browser ──HTTPS──▶ CloudFront ──┬─ viewer-request: Basic auth (CloudFront Function)
                                ├─ default behavior ──▶ S3 (private, OAC): index.html / app.js / styles.css
                                └─ /text-to-code ─────▶ Lambda Function URL (AWS_IAM + OAC): ttc-api-lambda
```

### `terraform/demo.tf` (new)

- **`ttc-api-lambda`** — same container image as the batch worker, with `image_config.command` overridden to `text_to_code_lambda.api_handler.handler` (the deployment shape #619 designed for). Models are already baked into the image, and both lambdas roll forward together on each deploy since they share `ttc_lambda_image_tag`. 3008 MB / 120s, OpenSearch-only IAM role (the API never touches S3).
- **Function URL locked to CloudFront** — `AWS_IAM` auth + a Lambda-origin Origin Access Control + `lambda:InvokeFunctionUrl` permission scoped to the distribution ARN. Direct calls to the URL get a 403.
- **Frontend bucket** — `dibbs-ttc-demo-frontend`, all public access blocked, readable only via CloudFront OAC. The three files deploy as `aws_s3_object` resources keyed on `filemd5`, and the distribution uses the managed `CachingDisabled` policy, so `terraform apply` is the whole deploy story — no sync scripts or invalidations.
- **Basic auth at the edge** — a `cloudfront-js-2.0` function on *both* behaviors (pages and API). After validating, it strips the viewer `Authorization` header so the OAC can attach its own SigV4 signature. Credentials come from `demo_auth_username` / `demo_auth_password`; the password has no default and is supplied by the `DEMO_AUTH_PASSWORD` GitHub secret (already created) via `TF_VAR_demo_auth_password` in `deploy.yaml`.
- The new lambda role is added to the OpenSearch domain access policy (in-place domain update; apply pauses on it for a few minutes).

### `frontend/app.js`

- `API_BASE` is now same-origin (`""`) when the page isn't served from localhost, so the deployed page calls the API through CloudFront with no CORS involved. The `demo.sh` local flow is unchanged (`127.0.0.1:8080`).
- Deployed POSTs include `x-amz-content-sha256`: CloudFront OAC signs Lambda-origin requests but doesn't compute POST payload hashes, so the client must supply it or Lambda rejects the request.
- API error messages on 502/504 now hint the service may be cold-starting (models load at init, ~10-30s).

## Verification

- `terraform fmt`/`validate` pass; `terraform plan` against live state shows **16 to add, 3 to change, 0 to destroy** — the changes are the OpenSearch access-policy statement plus image-tag diffs from local plan vars (CI supplies the SHA).
- After the first apply (~5-10 min for the distribution), from `terraform output demo_url`:
  - `curl -i <url>/` → 401 with `WWW-Authenticate: Basic`; with `-u dibbs:<pw>` → 200.
  - Authed `POST <url>/text-to-code` with the sha256 header → 200 with results; without the header → 403.
  - Direct Function URL / S3 object URLs → 403.
  - Browser: log in at the native prompt, run a single lookup and a CSV batch.

## Notes for reviewers

- First request after idle can 504 at CloudFront (60s origin cap vs. cold start); the 120s lambda timeout lets the container finish warming so a retry succeeds. Accepted for a demo — provisioned concurrency is the knob if it becomes annoying.
- The Basic auth token is embedded in the CloudFront Function code (visible in console/state). Fine for a shared demo credential; rotate by updating the secret and re-applying.


JNygaard-Skylight (approved):
You said this was good, so I'll take your word on it. Approved!

robertandremitchell (approved):
looks great!

PR #730, opened 2026-07-30
fix: Adding preprocessing of text candidate strings

### LINKED ISSUE #728: Clean up text before sending to OpenSearch
Original message: https://skylight-hq.slack.com/archives/C095WTSJR29/p1785434814163399

I've thought of a small preprocessing ticket I'd like to put into our code before our next APHL deploy. There have been a few patterns in the input data that are probably adding more noise than needed. The ticket is small, if someone wants to take it on while I do model tuning, but here's what we need to do (in the interest of creating more non-DS blocked work):

After we choose a text candidate to send to OpenSearch, but before we actually make the send request, we should perform the following three data cleaning operations:

Remove leading periods and colons from the input string (e.g. .AGE would become AGE, :rbc would become rbc). Trailing periods and colons should not be removed, as these can indicate information such as an abbreviation or the use of a fully-specified name.
Condense any consecutive space characters into a single space, and trim any spacing at the beginning and end of the input. For example, WBC  #  would become WBC # (going from 2 space characters to 1).  leuks would become leuks.
If the input is contained within an "arrow pattern," remove the arrows. A small number of inputs are sandwiched in the middle of a character demarcation like =>input<=. The => and <= are doing nothing for us, so we should remove them if the input is sandwiched like this. Equal signs and comparators that are not in this pattern should not be removed because they could carry meaningful semantic information.

robertandremitchell:
## Description

This adds a number of pre-processing cleanup recommendations from Brandon to the text candidate strings. Ideally, this will remove junk and also improve the result cache since there will be fewer text candidates that may differ on account of spacing or unneeded characters.

## Related Issues

Closes #728 

## Additional Notes

## Checklist

## Checklist for Reviewers


packages/text-to-code/src/text_to_code/services/eicr_processor.py
```diff
@@ -142,16 +142,40 @@ def _append_text_candidates_from_sub_node(
...
+        """
+        value = " ".join(value.split())
+
+        if value.startswith("=>") and value.endswith("<="):
+            value = value[2:-2].strip()
+
+        if value.startswith("<=") and value.endswith("=>"):
+            value = value[2:-2].strip()
+
+        return value.lstrip(".:")
```
nickclyde: I believe this can leave leading whitespace behind when a period or colon is followed by a space, which would undercut the cache-normalization goal. For example `". AGE"` becomes `" AGE"`, and `".: rbc"` becomes `" rbc"`, so `" rbc"` and `"rbc"` would still be distinct cache entries. Could we do `return value.lstrip(".: ")` (or add a final `.strip()`)? Might be worth adding a `(". AGE", "AGE")` case to the parametrized test too.

packages/text-to-code/tests/unit/test_eicr_processor.py
```diff
@@ -377,6 +377,102 @@ def test_extract_text_candidates_from_element_includes_child_tail_text(self):
...
+            ("<= input =>", "input"),
+            ("<= .AGE =>", "AGE"),
+            ("input.", "input."),
+            ("input:", "input:"),
+            ("WBC >= 10", "WBC >= 10"),
+            (">=input<=", ">=input<="),
+            ("=>input", "=>input"),
+            ("input<=", "input<="),
+            ("input=>value<=other", "input=>value<=other"),
+            ("<=input=>value<=other", "<=input=>value<=other"),
```
BradySkylight, line 406: We were actually seeing this in the data?  crazy!

BradySkylight (approved):
This is great.  It will be nice to add any other data cleaning 'tools' to this section over time!

nickclyde (commented):
Thanks for knocking this out so fast Rob! Just one small note:

packages/text-to-code/src/text_to_code/services/eicr_processor.py
BradySkylight: Do I dare say/suggest regexp?  :)

packages/text-to-code/tests/unit/test_eicr_processor.py
```diff
@@ -377,6 +377,102 @@ def test_extract_text_candidates_from_element_includes_child_tail_text(self):
...
+            ("=> input <=", "input"),
+            ("=> .AGE <=", "AGE"),
+            ("<=input=>", "input"),
+            ("<= input =>", "input"),
+            ("<= .AGE =>", "AGE"),
+            ("input.", "input."),
+            ("input:", "input:"),
+            ("WBC >= 10", "WBC >= 10"),
+            (">=input<=", ">=input<="),
+            ("=>input", "=>input"),
```
JNygaard-Skylight, line 403: We want to keep arrows if there is only one of them?

packages/text-to-code/src/text_to_code/services/eicr_processor.py
```diff
@@ -142,16 +142,40 @@ def _append_text_candidates_from_sub_node(
...
+        Current preprocessing steps:
+        1. Replace multiple whitespace characters with a single space and strip leading/trailing whitespace.
+        2. Remove leading and trailing 'arrows' if present ('=> ... <=' or '<= ... =>').
+        3. Remove leading periods and colons.
+
+        :returns: The preprocessed text candidate value.
+        """
+        value = " ".join(value.split())
+
+        if value.startswith("=>") and value.endswith("<="):
```
JNygaard-Skylight, line 171: Do we have to worry about pairs of arrows that not at the beginning and end? e.g. `"foo =>My test<= bar"`?

packages/text-to-code/tests/unit/test_eicr_processor.py
JNygaard-Skylight: Saw Brandon's initial Slack post, never mind!

packages/text-to-code/tests/unit/test_eicr_processor.py
robertandremitchell: according to brandon, yeah, wild stuff 😅

packages/text-to-code/src/text_to_code/services/eicr_processor.py
robertandremitchell: I don't think we have to worry about it, but I'm curious if this is a pattern @bamader has seen in the data?

packages/text-to-code/src/text_to_code/services/eicr_processor.py
robertandremitchell: we may need to eventually use regex, but most of what I've read on regex is that it is slower (https://medium.com/@conniezhou678/super-fast-string-matching-methods-vs-regular-expressions-in-financial-datasets-a-python-guide-67ab55db0f78). it's probably not really a massive delta, but since none of these were complex enough to need regex I figured it'd be better to avoid for now.

JNygaard-Skylight (approved):
Looks good to me!

packages/text-to-code/src/text_to_code/services/eicr_processor.py
BradySkylight: Yeah, I haven't used it a lot in Python, but in other eco-systems and it seems to do quite well.

nickclyde (approved):

PR #746, opened 2026-08-06
Refactor Terminology Update to use existing aws code in Lambda Handler Package

### LINKED ISSUE #731: Modify terminology_maintenance to leverage existing AWS/S3 functionality in lamba_handler.py

BradySkylight:
## Description

Refactor Terminology Update script to leverage existing code in the Lambda Handler package to access AWS S3 Bucket resources.  The changes are pretty straight forward, but if we need to do more I'm open to suggestions.

## Related Issues

Closes #731 

## Checklist for Reviewers


.github/scripts/terminology_maintenance.py
```diff
@@ -20,56 +18,15 @@
     set_loinc_response,
 )
 
+from lambda_handler.lambda_handler import create_s3_client, get_file_content_from_s3, put_file
```
JNygaard-Skylight, line 21: This is an annoying pedantic nit but,  if `lambda_handler` is now being used outside of the context of a lambda, it stops making sense to call it a lambda handler. I'm not sure if it makes sense to rename `lambda_handler` or pull the S3 stuff out of it? Or just leave it because it's literally fine, and I'm the only one that would be bothered by this.

BradySkylight: lol.  I mean you are technically correct.  I had pulled the S3 functionality out of it and into the Term. Maintenance for that reason, but it also makes sense to have the code live in one place.  I'm open to whatever, but we could always leave the code 'as is' and just remove this PR for now allowing the maintenance to use it's own functionality.  Thoughts?

JNygaard-Skylight: I think my desire to reduce  lines of code is greater than my dislike of slightly inaccurate names. I'm okay with it until someone comes up with a better solution.

robertandremitchell: yeah, probably worth a rename, but still overall right imo to keep as much AWS code in `lambda-handler` just to avoid deduplication or any versioning issues. maybe something broader like `setup-handler` or `utils`

robertandremitchell (approved):

JNygaard-Skylight (approved):
Approved!

